PR #27, opened 2021-06-05
Invocable actors

autodidaddict:
1. Removed the dependency on "lattice rpc". That crate can potentially even be deleted from the root of this repository.
2. Added a few time delays to a few of the tests because they were sporadically failing because we hadn't given things enough time to start or stop.
3. Added a test that invokes the echo actor with a constructed payload and asserts that we get the results we expect over NATS.
4. Made a couple of the tests `async: false` so that they can't run in parallel with other tests to avoid having counter problems (e.g. if we count the number of loaded actors, that can't be a concurrent thing while other tests are running)


host_core/native/hostcore_wasmcloud_native/src/inv.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /// The hash of the invocation's target, origin, and raw bytes
+    pub fn hash(&self) -> String {
+        invocation_hash(&self.target_url(), &self.origin_url(), &self.msg)
+    }
+
+    /// Validates the current invocation to ensure that the invocation claims have
+    /// not been forged, are not expired, etc
+    pub fn validate_antiforgery(&self) -> Result<()> {
```
brooksmtownsend, line 110: I think this is a good candidate for a Rust test here, to ensure this logic is working as intended as it's one of the more important pieces of auth in wasmcloud

host_core/test/host_core/actors_test.exs
```diff
@@ -20,4 +29,49 @@ defmodule HostCore.ActorsTest do
...
+      {:ok, %{body: body}} -> body
+      {:error, :timeout} -> :fail
+    end
+
+    assert res != :fail
+    HostCore.Actors.ActorSupervisor.terminate_actor(@echo_key, 1)
+
+    ir = res |> Msgpax.unpack!()
+
+    payload = ir["msg"] |> :binary.list_to_bin() |> Msgpax.unpack!()
```
brooksmtownsend, line 69: generic msgpack deserialization? HERESY!

host_core/test/host_core/providers_test.exs
```diff
@@ -15,11 +15,13 @@ defmodule HostCore.ProvidersTest do
         @httpserver_contract
       )
 
+    Process.sleep(1000)
     assert HostCore.Providers.ProviderSupervisor.all_providers() == [
              {@httpserver_key, @httpserver_link, @httpserver_contract}
            ]
 
     HostCore.Providers.ProviderSupervisor.terminate_provider(@httpserver_key, @httpserver_link)
+    Process.sleep(500)
```
brooksmtownsend, line 24: I'm ok with this for the time being, but I think we should write up an issue to be able to monitor the event stream or otherwise `wait` for certain events to happen. I'm imagining us writing up some code in the future to start an actor and provider, and to think it may be a race condition to invoke that actor/provider immediately feels bad

brooksmtownsend (changes requested):
Requested change is a rust test, others are comments

host_core/test/host_core/providers_test.exs
brooksmtownsend: Wrote up https://github.com/wasmCloud/wasmcloud-otp/issues/29

host_core/native/hostcore_wasmcloud_native/src/inv.rs
autodidaddict: :thinking: I was pretty sure there was a test for this in the "old" repo.. I'll see if I can find it.

host_core/test/host_core/providers_test.exs
autodidaddict: Agreed. Sleeps are a hack and a regression waiting to happen.

autodidaddict:
Added unit test for anti-forgery (copied from legacy repo)

brooksmtownsend (approved):
LGTM

PR #31, opened 2021-06-08
HTTPServer Go provider, e2e echo

autodidaddict:
some initial scaffolding that we @brooksmtownsend can use for when we pair on creating the preliminary http server provider.


autodidaddict (commented):
LGTM

brooksmtownsend (approved):
whoever wrote this code is a good coder

PR #32, opened 2021-06-10
Adding scaffold implementation of Redis K/V in Rust

autodidaddict:


capability-providers/keyvalue/rust/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+//!   * wasmbus.rpc.{prefix}.{public_key}.{link_name}.linkdefs.put - Puts a link def. Provider provisions resources for the given actor.
+//!   * wasmbus.rpc.{prefix}.{public_key}.{link_name}.shutdown - Request for graceful shutdown
+
+#[macro_use]
+extern crate log;
+#[macro_use]
+extern crate lazy_static;
+
+lazy_static! {
+    static ref LINKDEFS: RwLock<HashMap<String, LinkDefinition>> = RwLock::new(HashMap::new());
```
brooksmtownsend, line 18: Ha, look at how much work you need to make this accessible across the file! You should've written this in Go instead where you can do anything 🚗

capability-providers/keyvalue/rust/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[derive(Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Invocation {
+    pub origin: WasmCloudEntity,
+    pub target: WasmCloudEntity,
+    pub operation: String,
+    // may not need this since the host doesn't use it in the Rust invocation
+    //#[serde(with = "serde_bytes")]
+    pub msg: Vec<u8>,
```
brooksmtownsend, line 54: I think we did figure out we need this from our paired programming this AM, as without it the `msg` parameter will be a list of bytes instead of a byte blob

PR #35, opened 2021-06-10
Send termination signal to provider

autodidaddict:
Pay no attention to the branch name. I was going to refactor some params to keyword lists but the keyword list version was less safe/reliable than the regular one.


host_core/lib/host_core/providers/provider_supervisor.ex
```diff
@@ -34,6 +34,8 @@ defmodule HostCore.Providers.ProviderSupervisor do
     [{pid, _val}] = Registry.lookup(Registry.ProviderRegistry, {public_key, link_name})
     Logger.info("About to terminate child process")
     # TODO: send NATS shutdown message
+    prefix = Host.lattice_prefix()
+    :ok = Gnat.pub(:lattice_nats, "wasmbus.rpc.#{prefix}.#{public_key}.#{link_name}", "")
```
brooksmtownsend, line 38: If this is the shutdown message, it should be:
```
"wasmbus.rpc.#{prefix}.#{public_key}.#{link_name}.shutdown"
```

brooksmtownsend (changes requested):

brooksmtownsend (approved):

PR #41, opened 2021-06-11
Refactor and upgrade cache action

brooksmtownsend:
Currently the build pipeline doesn't successfully cache mix deps and dialyzer plts, which costs us about 8.5 minutes out of the 10 minutes of building and testing. This PR will _hopefully_ get that working.

Dependencies and PLTs will need to be rebuilt anytime we add new dependencies, so the first run afterwards may take the full 10 minutes.


brooksmtownsend:
Updated to allow dialyzer to finish with failures, because of https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action
> If the job completes successfully, the action creates a new cache with the contents of the path directory.

I'd like the job to complete successfully at least once, and I'll create an issue to fix the credo and dialyzer issues.

autodidaddict (approved):

PR #47, opened 2021-06-14
Adding RPC operations support to kv provider

autodidaddict:
PR commit prior to pairing on kvcounter actor work.

### DIFF
--- a/capability-providers/keyvalue/rust/src/generated.rs
+++ b/capability-providers/keyvalue/rust/src/generated.rs
@@ -1,0 +1,162 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct GetArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct AddArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+    #[serde(rename = "value")]
+    pub value: i32,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct SetArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+    #[serde(rename = "value")]
+    pub value: String,
+    #[serde(rename = "expires")]
+    pub expires: i32,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct DelArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct ClearArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct RangeArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+    #[serde(rename = "start")]
+    pub start: i32,
+    #[serde(rename = "stop")]
+    pub stop: i32,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct PushArgs {
+    #[serde(rename = "key")]
+    pub key: String,
+    #[serde(rename = "value")]
+    pub value: String,
+}
+
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct ListItemDeleteArgs {
+    #[serde(rename = "key")]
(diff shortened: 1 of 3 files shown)


capability-providers/keyvalue/rust/src/generated.rs
```diff
@@ -0,0 +1,162 @@
...
+    #[serde(rename = "value")]
+    pub value: String,
+}
+
+/// Response type for set add operations
+#[derive(Debug, PartialEq, Deserialize, Serialize, Default, Clone)]
+pub struct SetOperationResponse {
+    #[serde(rename = "new_count")]
+    pub new_count: i32,
+}
```
brooksmtownsend, line 155: This may have been an oversight in the previous version, but we should pick a variable casing guideline and stick to it. I know in the httpserver example we already use camelCase for `queryString`, and personally I'm more of a fan of camelCase, but I'm not against snake.

capability-providers/keyvalue/rust/src/main.rs
```diff
@@ -49,14 +53,46 @@ pub struct Invocation {
     pub origin: WasmCloudEntity,
     pub target: WasmCloudEntity,
     pub operation: String,
-    // may not need this since the host doesn't use it in the Rust invocation
-    //#[serde(with = "serde_bytes")]
+    // I believe we determined this is necessary to properly round trip the "bytes"
+    // type with Elixir so it doesn't treat it as a "list of u8s"
+    #[serde(with = "serde_bytes")]
```
brooksmtownsend, line 58: ✅

capability-providers/keyvalue/rust/src/main.rs
```diff
@@ -49,14 +53,46 @@ pub struct Invocation {
...
+            invocation_id: inv.id.to_string(),
+            ..Default::default()
+        }
+    }
+
+    /// Creates a successful invocation response. All invocation responses contain the
+    /// invocation ID to which they correlate
+    pub fn success(msg: impl Serialize) -> InvocationResponse {
+        InvocationResponse {
+            invocation_id: "".to_string(), // to be filled in later,
```
brooksmtownsend, line 89: I prefer all of my intentionally empty values to be something like `YEET` or `HEYOO`, but this works too.

capability-providers/keyvalue/rust/src/rpc.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(v) => v,
+        Err(e) => InvocationResponse::failure(&inv, &format!("{}", e)),
+    }
+}
+
+fn dispatch_operation(inv_id: &str, op: &str, actor: &str, bytes: &[u8]) -> OpResult {
+    let mut conn = actor_con(actor)?;
+    let inv = match op {
+        OP_ADD => add(&mut conn, deserialize(bytes)?),
+        OP_DEL => del(&mut conn, deserialize(bytes)?),
```
brooksmtownsend, line 39: Just a thought for codegen later, this feels like something we can absolutely generate scaffolding for. With the list of operations defined, go ahead and generate the long match and let people implement the individual functions.

capability-providers/keyvalue/rust/src/rpc.rs
```diff
@@ -0,0 +1,181 @@
...
+        exists: result,
+    }))
+}
+
+fn actor_con(actor: &str) -> Result<Connection, Box<dyn Error + Send + Sync>> {
+    let lock = crate::CLIENTS.read().unwrap();
+    if let Some(client) = lock.get(actor) {
+        client.get_connection().map_err(|e| format!("{}", e).into())
+    } else {
+        Err("No client for this actor. Did the host configure it?".into())
```
brooksmtownsend, line 179: We should add the actor ID in here for future debugging

brooksmtownsend (changes requested):
A few quick things, some stemming from the previous codegen I think.

capability-providers/keyvalue/rust/src/generated.rs
autodidaddict: Definitely could've been an oversight. These things were generated using older tools. This likely wouldn't/won't show up until an AssemblyScript actor goes to make a set operation. Once we're generating, things like this should be more consistent.

capability-providers/keyvalue/rust/src/main.rs
autodidaddict: having a const called `YEET` would let you write `YEET.into()`

brooksmtownsend: This is why I got into computer science.

brooksmtownsend (approved):

capability-providers/keyvalue/rust/src/generated.rs
brooksmtownsend: I'm for keeping with parity while we can with existing features 👍🏻

PR #48, opened 2021-06-14
Implemented call aliases, mostly

### LINKED ISSUE #4: Support for call aliases
Ensure that when an invocation is requested, it supports the use of a call alias if that alias is stored in the cache.

brooksmtownsend:
Intended to fix #4 

My idea here is to use an ets table to store call aliases. Whenever an actor starts, if it has a call alias, attempt to insert that public key into the ets table and subscribe to the call alias NATS topic to accept invocations at that topic.

For some reason here, the `:ets` table doesn't seem to be accessible properly inside of the actor module. I'm not sure if this is because the module cannot access the table, or if it's not writable, but as you can see the test that I prepared failed. Am I missing something obvious here @autodidaddict


brooksmtownsend:
Because some tests passed and some failed, I wonder if occasionally the :ets table is just not accessible in time.

autodidaddict:
Ok, there's a couple things going on here..

1. If we rely on a NATs subject for call alias invocation, then every time a client attempts to make an alias-based call to an alias that doesn't exist, then the client will sit and wait for the RPC timeout period before being told that there's likely no such call alias. In the current version of wasmCloud, when an invocation is created, we check if the target matches an actor, then a provider, then finally a call alias. If it's a call alias, we look up the alias and convert it to a public key. This allows for _fail fast_ scenarios, where the client will be informed immediately if there's no call alias.
2. ETS tables are, by default, private. They are only accessible to the actor that created them. Take a look at the lines of code found here: [https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/host.ex#L68](https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/host.ex#L68). 

For #2, what we want to do is create a `callalias_registry`, and then look up entries in that registry whenever an invocation is being resolved, and insert into that registry every time an actor is started with a call alias in their claims.

autodidaddict:
Another reason we don't want call aliases as NATS subjects is this limits the characters allowed. For example, if you use slashes in a NATS subject, it can potentially create conflicting or ineffective subscriptions.

brooksmtownsend:
@autodidaddict So I've got some improvements, I found where the other ETS tables are being setup, and I also found the `finish` function which seems to be the piece I was missing to understand where the actor/provider routing was happening. One thing I notice about the current code, is that there is no subscription to `wasmbus.rpc.#{prefix}.*`, so if we don't subscribe to the call alias then there's nothing to handle the invocation. I think if we want to be able to avoid that subscription, the actor supervisor should subscribe to that topic, but I like the current model how the actor itself is listening for invocations. Happy to chat about this when we pair next

brooksmtownsend:
There's not too much code in this PR, and a lot of it will change after #50, so I'm going to close it in favor of another one after #50 is merged

PR #55, opened 2021-06-17
Support for lattice prefix environment variable

autodidaddict:
You can now specify a lattice prefix via `WASMCLOUD_LATTICE_PREFIX`, with the default value being `default`


brooksmtownsend (approved):
if test happy, I happy

PR #56, opened 2021-06-17
Support for Call Alias

### LINKED ISSUE #4: Support for call aliases

brooksmtownsend:
Reopened issue from #48 

Fixes #4 

This PR also renames all `:ets` tables to "table", instead of "registry" to avoid confusion with the `Registry` module. 

This PR ALSO transforms the shutdown message publish into a `Gnats.req`, which allows the provider 2 seconds to clean up resources and respond. After the shorter length of time between 2 seconds and receiving a reply, the provider process is automatically terminated.


brooksmtownsend:
This appears to fail for the same reason as #55 , no unsubscribe logic for actors

host_core/lib/host_core/providers/provider_supervisor.ex
```diff
@@ -34,7 +34,17 @@ defmodule HostCore.Providers.ProviderSupervisor do
     [{pid, _val}] = Registry.lookup(Registry.ProviderRegistry, {public_key, link_name})
     Logger.info("About to terminate child process")
     prefix = HostCore.Host.lattice_prefix()
-    :ok = Gnat.pub(:lattice_nats, "wasmbus.rpc.#{prefix}.#{public_key}.#{link_name}.shutdown", "")
+    # Allow provider 2 seconds to clean up resources
+    case Gnat.request(
```
autodidaddict, line 38: Have the providers been modified to respond to this as a request? It used to be an async publish

autodidaddict (changes requested):
Just would like to see the capability providers modified so they're responding to the shutdown request.

host_core/lib/host_core/providers/provider_supervisor.ex
brooksmtownsend: Just modified the two providers we have, and I added a Makefile so we can easily rebuild them. It's not a perfect makefile but it'll do for now, especially when I have to swap them out for the Linux executables for committing.

Feasibly, we could have that part automated as well.

autodidaddict (approved):

Issue #43, opened 2021-06-11
Add OCI download capability to the native library


autodidaddict:
OCI file download (not necessarily the interpretation of said file) is currently available in here: https://github.com/wasmCloud/wasmCloud/blob/main/crates/wasmcloud-host/src/oci.rs

This issue will be considered closed when there is a _tested_ function in a NIF that can download an OCI file by reference and correctly verify its contents. This elixir-wrapped function will then be used as the basis for starting capability providers and actors by OCI reference instead of by supplying raw bytes.

autodidaddict:
Note that when an OCI reference (OCI URL -> public key) is established, it must be published on the appropriate NATS topic.
From our in-code docs:

* `wasmbus.rpc.{prefix}.ref_maps.put` - Publish a reference map, e.g. OCI ref -> PK, call alias -> PK
* `wasmbus.rpc.{prefix}.ref_maps.get` - Query all reference maps (queue subscribed by hosts)

PR #57, opened 2021-06-21
OCI functionality

autodidaddict:
Currently a WIP


brooksmtownsend (approved):
LGTM. Next we should add in a test to e2e to run through the kvcounter example or similar with only OCI references 👍🏻

autodidaddict:
Yep, to do that we'll need to add the actual functions to `ProviderSupervisor` and `ActorSupervisor` that take OCI references, which will include those things publishing on the `ref_map` subject, and we'll need to create the subscriber for ref maps to update the local cache the same way we do when we "hear" a claims put.

host_core/native/hostcore_wasmcloud_native/src/lib.rs
```diff
@@ -72,7 +72,7 @@ fn get_oci_bytes(
 ) -> Result<Vec<u8>, Error> {
     task::TOKIO
         .block_on(async { oci::fetch_oci_bytes(&oci_ref, allow_latest, allowed_insecure).await })
-        .map_err(|_e| rustler::Error::Atom("Failed to fetch OCI bytes"))
+        .map_err(|_e| rustler::Error::Term(Box::new("Failed to fetch OCI bytes")))
```
autodidaddict, line 75: Why the need for this change? Granted,  my use of `Atom` is just from copying and pasting from the various examples available for rustler.

autodidaddict:
Closes #43

host_core/native/hostcore_wasmcloud_native/src/lib.rs
brooksmtownsend: So previously, it would return the value `:"Failed to fetch OCI bytes"`, which wouldn't trigger the match for `{:error, reason}` inside of the `start_executable_from_oci` function. It was technically possible to match against that atom, which I'm fine with doing, but changing it to Term (from the Error docs):
```
    /// Encodes an arbitrary Boxed Encoder and returns `{:error, term}` from
    /// the NIF. Very useful for returning descriptive, context-full errors.
```
So `Term` seemed to return the proper tuple there

autodidaddict: Hah.. super good catch. We should make sure that all the other Atoms in this project do the same thing, and return the Error term, e.g. `{:error, "string"}` .. which is what I thought `error::Atom` did based on the samples I saw.

autodidaddict (commented):
Like the new refactor

host_core/test/host_core/wasmcloud/native_test.exs
```diff
@@ -18,18 +18,18 @@ defmodule HostCore.WasmCloud.NativeTest do
...
-          5_954_112
+          7_814_928
 
         {:unix, _linux} ->
-          9_734_792
+          7_911_908
 
         {:win32, :nt} ->
-          21_329_277
+          7_865_344
```
autodidaddict, line 32: Yay, windows support!

host_core/lib/host_core/claims/server.ex
```diff
@@ -0,0 +1,26 @@
...
+  # Topic wasmbus.rpc.{prefix}.claims.put, .get
+  # claims fields: call_alias, issuer, name, revision, tags, version, public_key
+
+  def request(%{topic: topic, body: body}) do
+    cmd = topic |> String.split(".") |> Enum.at(4)
+    Logger.info("Received claims command (#{cmd})")
+    # PUT
+    if cmd == "get" do
+      # {:reply, claims}
+    else
```
brooksmtownsend, line 14: Just a note to implement this when we need it, is there an issue for this? (#5 perhaps)

host_core/lib/host_core/providers/provider_supervisor.ex
```diff
@@ -25,6 +25,29 @@ defmodule HostCore.Providers.ProviderSupervisor do
...
+
+      bytes ->
+        par = HostCore.WasmCloud.Native.par_from_bytes(bytes |> IO.iodata_to_binary())
+
+        cache_path =
+          HostCore.WasmCloud.Native.par_cache_path(par.claims.public_key, par.claims.revision)
+
+        p = Path.split(cache_path)
+        tmpdir = p |> Enum.slice(0, length(p) - 1) |> Path.join()
+        IO.puts("TMPDIR #{tmpdir}")
```
brooksmtownsend, line 42: This should probably be a logger call, or removed

host_core/lib/linkdefs/server.ex
```diff
@@ -0,0 +1,23 @@
...
+    key = {ld["actor_id"], ld["contract_id"], ld["link_name"]}
+    map = %{values: ld["values"], provider_key: ld["provider_id"]}
+
+    Logger.info("Received link definition command (#{cmd})")
+
+    if cmd == "put" do
+      :ets.insert(:linkdef_table, {key, map})
+      :ok
+      # else if cmd == "get" do
+      # {:reply, linkdefs}
```
brooksmtownsend, line 17: same note as above to remember to implement 👍🏻

host_core/test/e2e/echo_test.exs
```diff
@@ -38,7 +38,7 @@ defmodule HostCore.E2E.EchoTest do
     Process.sleep(1000)
 
     :ok =
-      HostCore.LinkdefsManager.put_link_definition(
+      HostCore.Linkdefs.Manager.put_link_definition(
```
brooksmtownsend, line 41: I'm going to push a quick commit right after this review to change this in the dashboard as well.

brooksmtownsend (changes requested):
One tiny request, and something that I'm going to fix as well.

brooksmtownsend (approved):

PR #59, opened 2021-06-21
Implemented NATS retry logic

### LINKED ISSUE #22: Allow host to start before NATS
In an effort to keep things as simple and easy and _error-free_ as possible, we'd like to be able to start the host even if NATS isn't running. The host would essentially sit in an idle, non-functional, "not ready" state until it detects that NATS is available. Once NATS is available, it'll connect, make all its subscriptions, enable functionality, and put itself in the "ready" state.

brooksmtownsend:
Fixes #22

NATS can now be configured with the following environment variables:
```
WASMCLOUD_RPC_HOST
WASMCLOUD_RPC_PORT
WASMCLOUD_CTL_HOST
WASMCLOUD_CTL_PORT
```

Without it, the default value for host is `0.0.0.0` and port is `4222`. This PR also implements a naive retry logic that attempts to connect to NATS every second at the specified host/port value. This allows the user to start the host before configuring their NATS connection, which is essential for the edge case where NATS is started after the wasmcloud host.


host_core/lib/host_core/application.ex
```diff
@@ -26,4 +32,67 @@ defmodule HostCore.Application do
...
   end
+
+  defp retrieve_nats_env() do
+    rpc_host =
+      case System.get_env("WASMCLOUD_RPC_HOST") do
+        nil -> '0.0.0.0'
+        host -> host
+      end
+
+    rpc_port =
```
autodidaddict, line 43: You can supply default values to calls to `get_env` and avoid the case syntax entirely, see the [examples](https://hexdocs.pm/elixir/1.12/System.html#get_env/2-examples)

host_core/lib/host_core/application.ex
```diff
@@ -26,4 +32,67 @@ defmodule HostCore.Application do
...
+    case Gnat.start_link(%{host: rpc_host, port: rpc_port}, name: :lattice_nats) do
+      {:ok, _gnat} ->
+        :ok
+
+      {:error, :econnrefused} ->
+        Logger.warn(
+          "Failed to connect to lattice at #{rpc_host}:#{rpc_port}, retrying in 1 second..."
+        )
+
+        Process.sleep(1000)
```
autodidaddict, line 79: Rather than sleep, you should `send_after` which would then trigger a `handle_info` that responds to the message you sent to yourself. This is far more efficient (and concurrent-friendly) than sleeping the scheduler thread. Check out [send_after](https://hexdocs.pm/elixir/1.12/Process.html#send_after/3)

host_core/lib/host_core/application.ex
```diff
@@ -26,4 +32,67 @@ defmodule HostCore.Application do
...
+    case Gnat.start_link(%{host: ctl_host, port: ctl_port}, name: :control_nats) do
+      {:ok, _gnat} ->
+        :ok
+
+      {:error, :econnrefused} ->
+        Logger.warn(
+          "Failed to connect to control interface at #{ctl_host}:#{ctl_port}, retrying in 1 second..."
+        )
+
+        Process.sleep(1000)
```
autodidaddict, line 94: Same comment as above, you can do this in a `send_after` loop.

For bonus points, see about making the reconnect attempts do exponential backoff ;)

autodidaddict (approved):

autodidaddict (changes requested):
Per comments, there's some refactoring that can be done to clean things up

host_core/lib/host_core/application.ex
brooksmtownsend: Would this require refactoring the nats connection logic into a GenServer module? From some reading it looks like `handle_info` is a GenServer thing, not an Application thing (but I may be mistaken)

autodidaddict: yeah you generally want to keep all logic out of the `Application` entirely, and _just_ specify a list of children. I think an `Application` may actually be a `GenServer`, but it feels cleaner to have the connection be managed by a GenServer. We could probably put the `Gnat.connect` inside the `Host` GenServer or one of the other supervisors that automatically gets started.

autodidaddict: This will become a bit more tricky once we implement the lattice control interface, which will need to create subscriptions... so we'll have to ensure that those subscriptions aren't created until after the Gnat connection is managed. This leads me to think maybe we need a supervisor that tries to create a nats connection and then when it finally does, it spawns the `GnatServer` used for the request/reply features of the lattice control interface.

brooksmtownsend: I was doing some thinking about this, the original reason why I threw that logic in the Application was because there were a few children that make subscriptions, so I figured that there was no point in starting the children before Gnat was initialized. I am thinking something similar to your suggestion in your second comment:
> a supervisor that tries to create a nats connection and then when it finally does, it spawns the GnatServer used for the request/reply features of the lattice control interface
For example, a GnatSupervisor (name WIP) that handles creating the connection for the lattice / control interface, and then subscriptions are formed by sending a message to that supervisor with the PID of the interested GenServer.

For example, in `HostCore.LinkdefsManager`:
```
    {:ok, _sub} = Gnat.sub(:lattice_nats, self(), topic)
```
Would then change to something like
```
HostCore.GnatSupervisor.sub(:lattice_nats, self(), topic)
```

`HostCore.GnatSupervisor.sub/3` would accept a pid/atom of the connection, a pid of the interested genserver, and the topic. 

My biggest concern is ensuring that Gnat calls don't occur while we're waiting to establish the connection, I'm not sure if there's an option to start children in-order, let me know if I'm just missing that entirely

autodidaddict:
Let's chat about this because I think there's a larger issue and potentially a larger solution.

brooksmtownsend:
Closing for the time being, refer to the notes in #22

PR #60, opened 2021-06-22
Support different authentication modes for NATS

### LINKED ISSUE #7: Support environment variables for lattice configuration
Accept overridable NATS configuration data (e.g. JWT and seed authentication, configurable URL, etc) for both the lattice connection and the RPC connection.

brooksmtownsend:
Fixes #7 

Something about the code feels like I'm missing a cleaner way to do it (e.g. by filtering the connection options by `!= ""`) but it's not difficult to read and gets the job done.

FWIW I tested with user/pass, token, and anonymous. I did not setup a JWT/nk seed authentication challenge to test against, but that's a problem for tomorrow Brooks

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -37,7 +37,7 @@ defmodule HostCore.Host do
   def init(opts) do
     {host_key, host_seed} = HostCore.WasmCloud.Native.generate_key(:server)
 
-    start_gnat()
+    start_gnat(opts)
     configure_ets()
 
     Logger.info("Host #{host_key} started.")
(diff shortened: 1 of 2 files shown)


autodidaddict (approved):

host_core/lib/host_core/host.ex
```diff
@@ -68,9 +68,95 @@ defmodule HostCore.Host do
...
+      password: opts.ctl_pass,
+      token: opts.ctl_token,
+      nkey_seed: opts.ctl_seed,
+      jwt: opts.ctl_jwt
+    })
+  end
+
+  defp configure_lattice_gnat(opts) do
+    conn_settings =
+      Map.merge(%{host: opts.host, port: opts.port}, determine_auth_method(opts, "lattice"))
```
autodidaddict, line 95: There is no limit to the amount I love `Map.merge`

host_core/lib/host_core/host.ex
```diff
@@ -68,9 +68,95 @@ defmodule HostCore.Host do
...
+
+    case Gnat.start_link(conn_settings, name: :lattice_nats) do
+      {:ok, _gnat} ->
+        :ok
+
+      {:error, :econnrefused} ->
+        Logger.error("Unable to establish lattice NATS connection, connection refused")
+
+      {:error, _} ->
+        Logger.error("Authentication to lattice NATS connection failed")
```
autodidaddict, line 105: What's the return value of `Logger.error` ? Is it `:error` ? just trying to figure out what the return value of this function is when it fails and whether that corresponds to what `start_link`'s contract requires.

host_core/lib/host_core/host.ex
```diff
@@ -68,9 +68,95 @@ defmodule HostCore.Host do
...
+           jwt: jwt
+         },
+         conn_name
+       ) do
+    cond do
+      jwt != "" && nkey_seed != "" ->
+        Logger.info("Authenticating to #{conn_name} NATS with JWT and seed")
+        %{jwt: jwt, nkey_seed: nkey_seed, auth_required: true}
+
+      nkey_seed != "" ->
```
autodidaddict, line 143: This looks like an odd case. If we have a seed but we do _not_ have a JWT (what this predicate indicates), then we will get an authentication failure. User seed and JWT must always exist at the same time and we can't have one without the other.

autodidaddict (changes requested):
LGTM with the exception of that one edge case where it looks like you'll let people supply a seed without a JWT, which is guaranteed to produce an authentication failure with NATS IIRC.

host_core/lib/host_core/host.ex
brooksmtownsend: Oh gotcha, I was basing my cases off of the Gnat documentation here: https://hexdocs.pm/gnat/readme.html#authentication

You can see there's a case where just an nkey is accepted. If that's always going to be invalid, we can drop an error log here to handle that (or just remove the case entirely)

host_core/lib/host_core/host.ex
brooksmtownsend: re: https://hexdocs.pm/logger/1.12/Logger.html#error/2, it looks to return `:ok`, so it appears that `configure_lattice_gnat` will always return `:ok` unless an exception is thrown

host_core/lib/host_core/host.ex
autodidaddict: I stand corrected. It is in fact, entirely possible to authenticate using nothing but a user NKey and a challenge, as per [this section](https://docs.nats.io/nats-server/configuration/securing_nats/auth_intro/nkey_auth#generating-nkeys-and-configuring-the-server) of the NATS docs.

Carry on. As you were.

autodidaddict (approved):
LGTM

PR #61, opened 2021-06-24
Link dropdown autofill

brooksmtownsend:
This PR adds link def dropdown support for easier linking, as well as a few UI improvements.

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host_web/live/page_live.html.leex
+++ b/wasmcloud_host/lib/wasmcloud_host_web/live/page_live.html.leex
@@ -115,7 +115,6 @@
               <th>Instance Count</th>
               <th>Status</th>
               <th>Actor ID</th>
-              <%# <th>Action</th> %>
             </tr>
           </thead>
           <tbody>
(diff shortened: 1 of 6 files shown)


wasmcloud_host/lib/wasmcloud_host_web/live/page_live.html.leex
```diff
@@ -135,77 +136,91 @@
                     </svg>&nbsp;
                   </button>
                 </td>
-              </tr>
-            <% end %>
-          </tbody>
-        </table>
-      </div>
+                <!-- <td style="text-align: center">
```
autodidaddict, line 139: Can we remove the commented stuff?

autodidaddict (changes requested):

wasmcloud_host/lib/wasmcloud_host_web/live/page_live.html.leex
brooksmtownsend: Yeah that was leftover from some prototyping, I'll get it cleaned up

autodidaddict (approved):

PR #64, opened 2021-06-25
Bump to 0.20.0, bump wasmex and rustler versions

autodidaddict:


autodidaddict:
Ugh, we definitely need to figure out how to get rid of the non-determinism in that test that fails randomly, @brooksmtownsend

brooksmtownsend (approved):

PR #66, opened 2021-06-25
Adds the  function to the HostCore.Host

### LINKED ISSUE #20: Support host labels
Support the ability to specify custom host labels from the environment as well as the default/intrinsic labels like OS family, CPU, etc.

autodidaddict:
This will be used later for scheduling.

When starting iex (or the main application) with an environment variable like `HOST_TEST=23`, then we get the following host labels:
```
 HostCore.Host.host_labels
%{
  "hostcore.arch" => "x86_64",
  "hostcore.os" => "linux",
  "hostcore.osfamily" => "unix",
  "test" => "21"
}
```
Closes #20

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -42,15 +42,25 @@ defmodule HostCore.Host do
 
     Logger.info("Host #{host_key} started.")
 
-    # TODO - get namespace prefix from env/config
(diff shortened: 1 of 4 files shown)


brooksmtownsend (changes requested):
Only request, let's add a test that validates the proper host labels are added, and that the core host labels cannot be overwritten

brooksmtownsend (approved):

Issue #10, opened 2021-06-03
Perform health checks on actors and providers


autodidaddict:
Produce invocations that use the operation `HealthRequest` (or whatever the actual operation name is called) and periodically invoke actors and capability providers with this. Anything that fails a health request will emit a "health check failed" event. After a configurable number of failed events, the failing entity will be terminated.

PR #70, opened 2021-06-28
Health checks for providers and actors

autodidaddict:
Closing #10 

We now perform health checks every 30 seconds. For actors, the health check is performed internally and if we get any response at all from the actor, regardless of payload (this is status quo), then we assume it's healthy. We perform the same health check every 30 seconds for providers, except that we send the health check request payload directly on the providers `{rpc}.health` topic. If the provider fails to respond, we consider the health check a failure.

## Example Events

```
[#40] Received on [wasmbus.ctl.default.events]: '{"data":{"public_key":"MADQAFWOOOCZFDKYEYHC7AUQKDJTP32XUC5TDSMN4JLTDTU2WXBVPG4G"},"datacontenttype":"application/json","id":"8656a7a9-0727-464a-a60f-cda9216d2211","source":"NAVGJ2MWJYERFLQOKEDA57AFGB55YV6TPLHGBRRDSP75HVY2OCMOIRXK","specversion":"1.0","time":"2021-06-28T14:47:19.379813Z","type":"com.wasmcloud.lattice.health_check_passed"}'
[#43] Received on [wasmbus.ctl.default.events]: '{"data":{"link_name":"default","public_key":"VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M"},"datacontenttype":"application/json","id":"e9cb7cc6-b772-4caa-8fba-dc01fb91e321","source":"NAVGJ2MWJYERFLQOKEDA57AFGB55YV6TPLHGBRRDSP75HVY2OCMOIRXK","specversion":"1.0","time":"2021-06-28T14:47:57.518269Z","type":"com.wasmcloud.lattice.health_check_failed"}'
```


host_core/mix.exs
```diff
@@ -7,7 +7,7 @@ defmodule HostCore.MixProject do
       version: "0.20.0",
       elixir: "~> 1.12",
       start_permanent: Mix.env() == :prod,
-      compilers: [:rustler] ++ Mix.compilers(),
+      #   compilers: [:rustler] ++ Mix.compilers(),
```
brooksmtownsend, line 10: Is this an intentional change? I wasn't sure if this was left off to speed up dev time or something that should be removed entirely

host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -75,12 +79,87 @@ defmodule HostCore.Providers.ProviderModule do
...
+    {:noreply, state}
+  end
+
   def handle_info({_ref, msg}, state) do
     Logger.info(msg)
 
     {:noreply, state}
   end
 
+  defp publish_health_passed(state) do
```
brooksmtownsend, line 109: Feels like we _could_ refactor the publish_health_passed|failed here since it's mostly the same between actors and providers, but I'm less of a fan of creating a file to share between the two modules just for this purpose. Just a callout if we refactor it later.

brooksmtownsend (changes requested):
Just one request to check out the mix.exs change, otherwise LGTM

host_core/mix.exs
autodidaddict: Yes, this is intentional. We were getting a warning that the compiler was deprecated and the folks in the Rustler channel in Elixir slack told me to remove it.

brooksmtownsend (approved):

PR #72, opened 2021-06-28
Add OTP 24 to our build matrix

brooksmtownsend:


autodidaddict (approved):
We can probably take 22 out of here, and just support 23 and 24.

brooksmtownsend:
@autodidaddict OTP 22 is the default version that comes with the windows installer (when I installed elixir from the website with their `exe` installer, OTP 22 is the latest version available) so I think that should stay until the installer is updated

autodidaddict:
> @autodidaddict OTP 22 is the default version that comes with the windows installer (when I installed elixir from the website with their `exe` installer, OTP 22 is the latest version available) so I think that should stay until the installer is updated

Bloody windows. Always ruining everything.

Issue #67, opened 2021-06-25
Create a distillery release


autodidaddict:
Add distillery to the project and run through the process of manually creating a release. Automate this into the CI pipeline so that when a `v0.x.y` tag is created, a new distillery release is created for Linux, Mac, and Windows and attached to the release as a `.tar.gz` file, e.g. `wasmcloud-host-x86_linux.tar.gz`.

Let's also figure out a way to automate the generation of the `aarch64` image.

brooksmtownsend:
### Finished items
- [x] Automation of creating the `wasmcloud_host-v0.20.0-x86_64-linux-gnu.tar.gz` artifact
- [x] Automation of creating and pushing the `wasmcloud_host:0.20.0` Docker image (hosted on wasmcloud.azurecr.io and docker.io)

### In Progress
- [ ] Automation of creating the `wasmcloud_host-v0.20.0-x86_64-macos.tar.gz` artifact using macos github runner

### Need to be done
- [ ] Automate the creation of release with a pushed tag that includes the above artifacts
- [ ] Automation of creating the `wasmcloud_host-v0.20.0-x86_64-windows.tar.gz` artifact
- [ ] Automation of creating the `wasmcloud_host-v0.20.0-aarch64-linux-gnu.tar.gz` artifact

### Optional / Post MVP
- [ ] Automation of creating the `wasmcloud_host-v0.20.0-aarch64-macos.tar.gz` artifact

@billyoung I'd be happy to pull up for a quick chat to go over the work that I've done, but the TL;DR is that the elixir runtime is packaged with a [distillery release](https://hexdocs.pm/distillery/home.html). This creates a standalone `.tar.gz` which can be unzipped and then ran using `./install_dir/wasmcloud_host foreground` or `./install_dir/wasmcloud_host start`. This, by itself, is an awesome system, but there are 2 parts that make it difficult.

1. The elixir runtime system (ERTS) is bundled with the application by default. This is compiled for an operating system/architecture, so it has to match. We have the option to not include the elixir runtime, but if we do that then the user must install the exact same version of elixir and erlang that we built it with or the project will fail to run.
2. We use Native Implemented Functions (NIFs) to bring in some of our Rust functionality as callable C functions. These must be compiled for an OS/ARCH, so even if we don't bundle the ERTS the NIFs will still have to match the host machine.

[...]

PR #86, opened 2021-07-01
Distillery Release

brooksmtownsend:
Part of #67 

## MVP Updated PR description
As mentioned in #67, the following parts have been completed:
-  Automation of creating the wasmcloud_host-v0.20.0-x86_64-linux-gnu.tar.gz artifact
-  Automation of creating and pushing the wasmcloud_host:0.20.0 Docker image (hosted on wasmcloud.azurecr.io and docker.io)
-  Automation of creating the wasmcloud_host-v0.20.0-x86_64-macos.tar.gz artifact using macos github runner

### Original Text
WIP with github actions to create and distribute a distillery release when a github release is created

Laying out some thoughts here, my ideal end goal is when we create a github release is that we release versions of wasmcloud_host (the web UI + host) and host_core (only the host, minimal) in the following build matrix:
os: [linux, macos, windows]
arch: [x86_64, aarch64]

We should produce 2 different artifacts for each combination. One that includes the web UI, and the `minimal` artifact that simply launch the host without the overhead of the phoenix server or the web assets (that are frankly a bit large)
- `wasmcloud_host-v<VERSION>-<os>-<arch>.tar.gz`
- `wasmcloud_host-minimal-v<VERSION>-<os>-<arch>.tar.gz`

**EDIT**
Currently, I have the ability to build a docker image including wasmcloud_host and two of the release artifacts we're interested in:
- `wasmcloud_host-v0.20.0-linux-x86_64-musl.tar.gz`
- `wasmcloud_host-v0.20.0-linux-x86_64-gnu.tar.gz`

We may not want to distinguish between libc installations, but it's worth mentioning that there are some compatibility issues between the two because of the Rust NIFs (musl libc vs glibc). 

The next lift of finishing out the release versions is _hopefully_ not going to be too bad, but compiling for windows, macos, and aarch64 are still unknowns.


autodidaddict:
so, just to clarify:
* `wasmcloud_host-v{VERSION}-<os>-<arch>.tar.gz` will be a distillery release for `wasmcloud_host`
* `wasmcloud_host-minimal-v{VERSION}-<os>-<arch>.tar.gz` will be a distillery release for `host_core`
?

brooksmtownsend:
@autodidaddict  that's correct. The goal of the minimal release would be to release the smallest tarGEEzy with all of the host functionality

brooksmtownsend:
Another option for build releases is not including the elixir runtime in some of our releases (simple flag `include_erts: false`). This doesn't bundle the elixir runtime with the release, which means that the user must install the same version of elixir, otp, and from my testing they must have the appropriate gcc installation for our Rust NIFs.

The idea for this would be that it's a release like `wasmcloud_host-v0.20.0-elixir1.12.2-otp24.tar.gz` and we'd need to include some installation instructions for the NIF required things. The benefit is that we'd support any architecture/os combo that elixir/otp/gcc supports.

brooksmtownsend:
> The downside is then the burden of installing the gaps in the matrix falls on the user instead of us. We need to keep the dev experience primary.

The more I think about it as well, is that if the user is going to go through all of those steps they might as well just clone the repository and run the server directly and not have to deal with matching elixir/otp versions.

autodidaddict:
Yeah if you want to mix and match and control the matrix yourself,  might as well do the build yourself. Releases should be for the easy things like "give me the OTP 24 version of wasmcloud" or "give me the OTP 23 version" etc.

autodidaddict:
This is a staggering amount of work... :clap: :clap: well done.

autodidaddict (approved):

brooksmtownsend:
@billyoung do you have any issues with anything here?

billyoung:
@brooksmtownsend looks great! :shipit:

PR #88, opened 2021-07-06
Renamed public_key field to sub for claims

brooksmtownsend:
This was causing a bug with the claims when starting an actor where we were looking for `public_key` and instead needed to look for the claims field `sub`.

@stevelr This is a small fix but just make sure that it works for your testing 👍🏻


stevelr:
I'll test with that - Thank!

stevelr (approved):

PR #97, opened 2021-07-13
Topic fixes

autodidaddict:
Fixing provider health check topic, moving events topic to `wasmbus.evt.(prefix)`

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
+++ b/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
@@ -20,7 +20,7 @@ defmodule WasmcloudHost.Lattice.StateMonitor do
     state = %State{actors: %{}, providers: %{}, linkdefs: %{}, refmaps: %{}, claims: %{}}
     prefix = HostCore.Host.lattice_prefix()
 
-    topic = "wasmbus.ctl.#{prefix}.events"
+    topic = "wasmbus.evt.#{prefix}"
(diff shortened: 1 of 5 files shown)


stevelr (dismissed):
these look fine.

wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
```diff
@@ -61,7 +61,7 @@ defmodule WasmcloudHost.Lattice.StateMonitor do
 
     state =
       cond do
-        String.ends_with?(topic, ".events") ->
+        String.contains?(topic, ".evt") ->
```
brooksmtownsend, line 64: Should this be `.evt.` to ensure it's in the middle? Or perhaps split by period and compare the second value to `evt`?

brooksmtownsend (changes requested):

wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
brooksmtownsend: Just wouldn't want someone's lattice prefix to be `evt` and trigger this.

brooksmtownsend (approved):

wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
brooksmtownsend: That looks good, I get wary with `contains` since that can pretty easily match on unintended strings. Change LGTM

PR #98, opened 2021-07-13
Cluster seed identity.

autodidaddict:


autodidaddict:
This PR does **not** include validation of invocations coming from capability providers because we simply don't have any providers that are creating invocations with embedded claims tokens. Once we have those, then we can simply add the 2-3 lines of code that verifies the invocation.

All outbound invocations coming from actors are now being signed by the _cluster seed_ and not the host key.
There are now environment variables being used for both the cluster seed and the list of valid issuers. Once we start enforcing inbound invocations, we'll be enforcing the check against the valid issuers as well.

autodidaddict:
Created a follow-up issue so that we can track enabling the AFT validation on provider->actor calls. #99

host_core/test/host_core/wasmcloud/native_test.exs
```diff
@@ -67,11 +67,49 @@ defmodule HostCore.WasmCloud.NativeTest do
...
+        @httpserver_key,
+        @httpserver_contract,
+        @httpserver_link,
+        "HandleRequest",
+        req
+      )
+
+    res =
+      HostCore.WasmCloud.Native.validate_antiforgery(inv |> IO.iodata_to_binary(), [
+        "CSUPERBADKEYIAMAMALICIOUSACTOR"
```
brooksmtownsend, line 108: Love it

host_core/lib/host_core/host.ex
```diff
@@ -204,7 +212,8 @@ defmodule HostCore.Host do
       lattice_rpc_url: "",
       provider_key: provider_key,
       # TODO
-      env_values: %{}
+      env_values: %{},
+      invocation_seed: cluster_seed()
```
brooksmtownsend, line 216: Having a little bit of trouble tracking this and feeling good about it, so here for the hostinfo we are passing the cluster seed key over to the providers. Is the cluster seed a private key that we are okay handing out to the providers? If (when?) we implement provider discovery, wouldn't any provider that has access to the lattice also have access to the cluster key anyways?

Let me know if this cluster seed is the public keypair, or if I'm thinking about what the cluster should be incorrectly.

brooksmtownsend (changes requested):

host_core/lib/host_core/host.ex
autodidaddict: In this situation we are passing the cluster seed to the provider because this provider is _trusted_ by virtue of us a) having access to the PAR file and b) being the ones who started it in the first place. So, we give it a seed (private) key so that it can sign invocations bound for any host that has the public half of that in the list of valid issuers.

We don't trust external (not even implemented yet) providers, so we will _not_ give them a signing key - they'll have to be given one by whatever process started them in the first place. Access to the signing key _without the host runtime providing it_ is essentially what guarantees the authenticity of the remote/external providers.

autodidaddict: I don't feel like this is the best solution, but it's **a** solution right now. If we are going to protect invocations by requiring them to be signed by one of the valid/known issuers, then we have to give the providers we auto-launch **a** signing key, whereas external providers we assume they've gained access to a key because they deserve it.

In an ideal world there would be a simpler way to prevent unauthorized access to the lattice once inside a NATS connection (defense in depth), but I don't yet know of a simpler way than signing invocations.

autodidaddict: this makes a good argument for (eventually when supported) using an OPA policy that allow-lists certain provider issuers because if you allow _any_ issuer, then anyone could put up a bogus provider in a public registry and if they gained access to the control interface, they could request the starting of the trojan PARjeezy, which could then exfiltrate the current signing key.

This is still pretty far-remote in terms of likelihood because for this attack vector to work, the attacker needs to compromise **three** things - NATS access to the control interface, NATS access to the RPC interface, and publish a trojan capability provider to a registry that the environment has access to.

brooksmtownsend: Okay, I'm fine with it for now then, I know we discussed the cluster signing as a mechanism for defense in depth so I just wanted to understand the implementation 👍🏻 

We'll just want people to keep in mind that if they are starting a provider in their lattice, that you have to trust it (OPA will be good for this) and it's not the same level of security that WASM guarantees re: not accessing information it's not supposed to

brooksmtownsend (dismissed):

autodidaddict:
Per discussion : going to move the `WASMCLOUD_HOST_DATA` environment variable to a `host_data.json` file that is written to the provider's launch directory prior to execution. This prevents the attack vector of someone with `ps -e` access being able to read environment variables.

autodidaddict:
@brooksmtownsend @stevelr the latest commit to this has the host data structure being sent via `stdin`

host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -41,7 +41,8 @@ defmodule HostCore.Providers.ProviderModule do
       |> Base.encode64()
       |> to_charlist()
 
-    port = Port.open({:spawn, "#{path}"}, [:binary, {:env, [{'WASMCLOUD_HOST_DATA', host_info}]}])
+    port = Port.open({:spawn, "#{path}"}, [:binary])
+    Port.command(port, "#{host_info}\n")
```
brooksmtownsend, line 45: Worth noting we can still use environment variables for providers down the line if we want to make any other non-sensitive information available

capability-providers/keyvalue/rust/src/main.rs
```diff
@@ -112,7 +114,9 @@ fn main() -> Result<()> {
     let _ = env_logger::try_init();
 
     // Host data is a JSON object that is base64 encoded
-    let host_data_enc = std::env::var("WASMCLOUD_HOST_DATA")?;
+    let stdin = io::stdin();
+    let host_data_enc = stdin.lock().lines().next().unwrap().unwrap();
+    //let host_data_enc = std::env::var("WASMCLOUD_HOST_DATA")?;
```
brooksmtownsend, line 119: This comment can be removed now

capability-providers/keyvalue/rust/src/main.rs
```diff
@@ -112,7 +114,9 @@ fn main() -> Result<()> {
     let _ = env_logger::try_init();
 
     // Host data is a JSON object that is base64 encoded
-    let host_data_enc = std::env::var("WASMCLOUD_HOST_DATA")?;
+    let stdin = io::stdin();
+    let host_data_enc = stdin.lock().lines().next().unwrap().unwrap();
```
brooksmtownsend, line 118: I know this is still prototype provider code, but I would feel better without the double unwrap if it was something like:
```
let stdin = io::stdin();
let mut handle = stdin.lock();
let host_data_enc = String::new();
handle.read_to_string(&mut host_data_enc)?;
```

Just so that the error bubbles up instead of panicking, and since we are only reading one line. Thoughts?

brooksmtownsend (changes requested):
only a request to

capability-providers/keyvalue/rust/src/main.rs
brooksmtownsend: (I ripped that example from https://doc.rust-lang.org/std/io/struct.Stdin.html#examples-1). According to https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_line it reads until the newline byte

brooksmtownsend (approved):

PR #123, opened 2021-07-18
capability provider now gets its own unique NATS credential

autodidaddict:
We've encoded an opinion in this PR that we no longer want to support centralized authentication. As such, the environment variables that you can pass to the wasmcloud host now include just the `JWT` and `SEED` which together represent either JWT+seed auth or just seed auth.

Traditional username/pasword and token are no longer allowed because they represent centralized authentication and a configuration burden on operations that we don't want to inflict on wasmCloud users.


autodidaddict:
There is now a completely separate set of environment variables that will be passed along to managed capability providers to be used for their connection to the lattice. this is being done _instead of_ minting brand new credentials on demand at start time.

brooksmtownsend (approved):

PR #131, opened 2021-07-20
Adding a purge function to HostCore.Host that empties actors and providers gracefully

autodidaddict:

### DIFF
--- a/host_core/test/host_core_test.exs
+++ b/host_core/test/host_core_test.exs
@@ -2,6 +2,12 @@ defmodule HostCoreTest do
   use ExUnit.Case
   doctest HostCore
 
+  @httpserver_path "priv/providers/httpserver"
+  @echo_path "priv/actors/echo_s.wasm"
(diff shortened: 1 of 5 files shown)


host_core/test/host_core_test.exs
```diff
@@ -23,4 +29,29 @@ defmodule HostCoreTest do
     # HOST_ prefix removed.
     assert "42" == labels["testing"]
   end
+
+  test "Host purges actors and providers" do
+    {:ok, bytes} = File.read("priv/actors/echo_s.wasm")
+    {:ok, _pid} = HostCore.Actors.ActorSupervisor.start_actor(bytes)
+
+    {:ok, _pid} =
```
brooksmtownsend, line 37: This test should also have the `on_exit` calls that are present in other tests just to ensure that the actor and provider are cleaned up at the end

brooksmtownsend (changes requested):
Just a quick note to prevent testing errors

host_core/test/host_core_test.exs
autodidaddict: I actually left those off because the test itself is testing the ability to purge the host of actors and providers. If the test passes, there are no more providers or actors :)

brooksmtownsend (approved):

PR #133, opened 2021-07-21
Query lattice cache on startup

### LINKED ISSUE #63: Host should query distributed cache upon startup
Upon startup, a host must make requests of the appropriate NATS topics so that it can refresh the cache (ETS tables) of the following types:
* Claims
* Link Definitions
* Reference Maps / OCI References

brooksmtownsend:
Fixes #63 

This PR implements the Host querying the lattice cache upon startup. This means that when a Host is initialized, if at least one other host exists on the same lattice, it will request the following information from that host:
- OCI references
- Call aliases
- Actor claims
- Link definitions

The newly-started host will then populate its local cache, and subscribe to the same topics as the other hosts. This means that hosts can now join the lattice later and still have access to the same local information as a previously started host (this was accomplished with the distributed cache provider in `0.18.0`)

If the host is the first to join the lattice, the requests will simply timeout after 2 seconds and the caches will start empty as expected.

### Extra stuff
- Updated the version to `0.50.0` for the in da OTP release
- Updated the release actions to automatically source the version for the docker image

Signed-off-by: Brooks Townsend <brooksmtownsend@gmail.com>


host_core/lib/host_core/host.ex
```diff
@@ -47,7 +47,48 @@ defmodule HostCore.Host do
...
+     }, {:continue, :query_lattice_cache}}
+  end
+
+  # TODO: Evaluate downsides of a potential race condition here. May be better to simply execute this as part of the init function
+  @impl true
+  def handle_continue(:query_lattice_cache, state) do
+    Logger.info("Querying lattice for cache...")
+
+    results =
+      ParallelTask.new()
```
autodidaddict, line 59: I <3 this paralleltask

host_core/test/e2e/controlinterface_test.exs
```diff
@@ -0,0 +1,7 @@
+defmodule HostCore.E2E.ControlInterfaceTest do
+  use ExUnit.Case, async: false
+
+  test "it works" do
+    assert true
```
autodidaddict, line 5: Tests pass. Ship it! :rocket: :shipit:

host_core/lib/host_core/host.ex
brooksmtownsend: It's a pretty sweet package

brooksmtownsend: And it's really nice here because we have a few queries that are okay to timeout, but we don't necessarily want to wait sequentially for them to timeout

autodidaddict (approved):

PR #134, opened 2021-07-21
Implement the built-in logging provider

autodidaddict:

### DIFF
--- a/host_core/lib/host_core/web_assembly/imports.ex
+++ b/host_core/lib/host_core/web_assembly/imports.ex
@@ -121,63 +121,134 @@ defmodule HostCore.WebAssembly.Imports do
     claims = state.claims
     actor = claims.public_key
 
-    provider_key =
-      case HostCore.Providers.lookup_provider(namespace, binding) do
-        {:ok, pk} -> pk
-        :error -> ""
-      end
-
-    case HostCore.Linkdefs.Manager.lookup_link_definition(actor, namespace, binding) do
-      # Authorize actor to invoke provider over link definition
-      {:ok, _ld} ->
-        authorize(
-          actor,
-          binding,
-          namespace,
-          operation,
-          payload,
-          claims,
-          seed,
-          prefix,
-          provider_key,
-          state
-        )
-        |> finish_host_call(agent)
-
-      # Link definition not found for actor, could be attempting to call a call alias
-      :error ->
-        case lookup_call_alias(namespace) do
-          {:ok, actor_key} ->
-            finish_host_call(
-              {:ok, actor_key, binding, namespace, operation, payload, claims, seed, prefix,
-               provider_key, state},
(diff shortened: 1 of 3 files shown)


autodidaddict:
@brooksmtownsend looking for some feedback on the invocation refactor before I move on to adding in support for the built-in logging provider. We're going to repeat this pattern for other built-in providers so I want to get it clean first.

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -121,132 +121,148 @@ defmodule HostCore.WebAssembly.Imports do
...
     0
   end
 
-  defp finish_host_call(
-         {:ok, actor, binding, namespace, operation, payload, _claims, seed, prefix, provider_key,
-          state},
-         agent
+  defp invoke(
+         _token = %{
+           authorized: true,
```
brooksmtownsend, line 220: I struggled with this when refactoring, is there any variation of function names that makes `invoke` and `do_invoke` less confusing? I like this better than `finish_host_call` and `invoke`, 🤔 

Thoughts on the top-level pipeline to be:
```
token
    |> identify_target()
    |> verify_link()
    |> authorize_call()
    |> generate_invocation()
    |> invoke()
```

Just a brainstorm. If this feels worse to you then I'm happy with where it is

autodidaddict: refactored the names here

autodidaddict:
@brooksmtownsend pushed the initial implementation of the builtin logging provider. I'm unsure if there's a good testing strategy here until we've got some new actors that actually use logging. Even then, not sure we'll be able to do anything like reliably capture stdout to ensure that logging is happening other than just eyeballing it when we run a manual test.

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -2,6 +2,8 @@ defmodule HostCore.WebAssembly.Imports do
   require Logger
   alias HostCore.Actors.ActorModule.State
 
+  @wasmcloud_logging "wasmcloud:builtin:logging"
+
```
brooksmtownsend, line 6: we already mentioned it in slack but we just have to remember to change wash and tell others to re-sign

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -121,132 +123,166 @@ defmodule HostCore.WebAssembly.Imports do
...
+        seed,
+        actor,
+        target_type,
+        target_key,
+        namespace,
+        binding,
+        operation,
+        payload
+      )
+      |> perform_rpc_invoke(target_subject)
```
brooksmtownsend, line 262: I like this 👍🏻

brooksmtownsend (approved):

Issue #15, opened 2021-06-03
Implement the built-in "extras" provider


autodidaddict:
There's a subtle, but important point with this one. If we implement the extras provider as a standard subscriber to NATS topics, then we'll basically have one extras provider per host, and they'll each respond to all of the requests of the extras provider.

On the other hand, if the extras provider subscribes as a queue subscriber to the specific subject for it, and registers itself during startup, then _an_ extras provider will be randomly chosen from among the hosts in the lattice. Given the way NATS works, we can at least guarantee that the one chosen will be in an optimal path, but will never be local. 

In short, the extras provider will incur a network latency penalty just as if it were a "regular" capability provider that was registered and managed externally.

Another alternative is to monitor the namespace and operation for each invocation, and if we see the extras provider, then we can defer those specific invocations to a custom provider.

I prefer the former, where we create a provider that's queue subscribed to the appropriate topic. This will ensure that all provider invocations go through the same code path and we don't end up with all kinds of one-off conditional branches like we had with Actix and the version before it.

stevelr:
depends on issue #114

autodidaddict:
Per multiple discussions, we are removing the sequence number generation because of its local nature potentially causing unintended problems as actors scale horizontally (each host would produce its own sequences with potential overlap).

Additionally, we've also decided to rename this contract ID to `wasmcloud:builtin:numbergen`.

PR #139, opened 2021-07-23
Initial commit of numbergen builtin provider

autodidaddict:
Closing #15

### DIFF
--- a/host_core/lib/host_core/providers/builtin/numbergen.ex
+++ b/host_core/lib/host_core/providers/builtin/numbergen.ex
@@ -1,0 +1,17 @@
+defmodule HostCore.Providers.Builtin.Numbergen do
+  def invoke("GenerateGuid", _payload) do
+    Msgpax.pack!(UUID.uuid4())
+  end
+
+  def invoke("RandomInRange", payload) do
+    params = Msgpax.unpack!(payload)
+
+    min = min(params["min"], 0)
+    max = max(params["max"], 4_294_967_295)
+    Msgpax.pack!(Enum.random(min..max))
+  end
+
+  def invoke("Random32", _payload) do
+    Msgpax.pack!(floor(:rand.uniform(4_294_967_294)))
+  end
+end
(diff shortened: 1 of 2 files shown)


autodidaddict:
This is based on the data structures and operation names from the pending PR to the smithy-based actor interface for `wasmcloud:builtin:numbergen`

host_core/lib/host_core/providers/builtin/numbergen.ex
```diff
@@ -0,0 +1,17 @@
+defmodule HostCore.Providers.Builtin.Numbergen do
+  def invoke("GenerateGuid", _payload) do
+    Msgpax.pack!(UUID.uuid4())
+  end
+
+  def invoke("RandomInRange", payload) do
```
brooksmtownsend, line 6: I like this specification of ambiguous things (inclusive range vs exclusive) in the operation name for clarity

host_core/lib/host_core/providers/builtin/numbergen.ex
```diff
@@ -0,0 +1,17 @@
...
+    params = Msgpax.unpack!(payload)
+
+    min = min(params["min"], 0)
+    max = max(params["max"], 4_294_967_295)
+    Msgpax.pack!(Enum.random(min..max))
+  end
+
+  def invoke("Random32", _payload) do
+    Msgpax.pack!(floor(:rand.uniform(4_294_967_294)))
```
brooksmtownsend, line 15: What number does `4294967294` correspond to? Is it the largest 32-bit integer without a signed bit? If so, some languages use signed integers and may not support this big of a number if they try to store it in an int32

brooksmtownsend (changes requested):

host_core/lib/host_core/providers/builtin/numbergen.ex
autodidaddict: It's the largest U32. The smithy spec for this value calls for it to be a u32 so I'm basically just enforcing that it can be no bigger than the biggest U32.

brooksmtownsend (approved):

brooksmtownsend:
If a language only supports I32s, then they can request a random within range of that language's `0` and `INT32::MAX`

PR #140, opened 2021-07-23
Configurable delay between provider termination request and halt

autodidaddict:


brooksmtownsend (approved):
This is nice. I like leaving 2s as the default, some additional argument there is that AWS Lambda gives your function 2s to clean up and SIGKILLs it after that.

PR #142, opened 2021-07-27
Adding support for live updates

autodidaddict:


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -55,11 +55,47 @@ defmodule HostCore.Actors.ActorModule do
     GenServer.call(pid, :health_check)
   end
 
+  def live_update(pid, bytes, claims) do
+    GenServer.call(pid, {:live_update, bytes, claims}, 30_000)
```
brooksmtownsend, line 59: We can re-use the `@thirty_seconds` constant here

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -55,11 +55,47 @@ defmodule HostCore.Actors.ActorModule do
...
+    imports = %{
+      wapc: Imports.wapc_imports(agent),
+      wasmbus: Imports.wasmbus_imports(agent)
+    }
+
+    # shut down the previous Wasmex instance to avoid orphaning it
+    old_instance = Agent.get(agent, fn content -> content.instance end)
+    GenServer.stop(old_instance, :normal)
+
+    {:ok, instance} = Wasmex.start_link(%{bytes: bytes, imports: imports})
```
brooksmtownsend, line 79: Doing some thinking here, so each ActorModule will receive the call individually, so if I have 100 instances of actor `MASDASDASD`, there will be 100 info logs saying that actor is performing a live update?

host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -38,11 +38,44 @@ defmodule HostCore.Actors.ActorSupervisor do
...
         start_actor(bytes |> IO.iodata_to_binary(), oci)
     end
   end
 
+  def live_update(oci) do
+    with {:ok, bytes} <- HostCore.WasmCloud.Native.get_oci_bytes(oci, false, []),
+         {:ok, new_claims} <-
+           HostCore.WasmCloud.Native.extract_claims(bytes |> IO.iodata_to_binary()),
+         {:ok, old_claims} <- HostCore.Claims.Manager.lookup_claims(new_claims.public_key),
+         :ok <- validate_actor_for_update(old_claims, new_claims) do
```
brooksmtownsend, line 51: I've never seen this syntax before, is this just a bunch of async calls? Do they happen in sequence?

brooksmtownsend: It reads like "once I've done these operations in this order, do the following code", just making sure I understand properly

host_core/lib/host_core/actors/actor_module.ex
brooksmtownsend: I think this might be better as either a trace log, or placed in the ActorSupervisor `live_update` section

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -157,11 +157,18 @@ defmodule HostCore.ControlInterface.Server do
...
     if host_id == HostCore.Host.host_key() do
-      _update_actor_command = Jason.decode!(body)
+      update_actor_command = Jason.decode!(body)
 
-      # TODO - live updates for actors is not implemented yet
+      response =
+        case HostCore.Actors.ActorSupervisor.live_update(update_actor_command["new_actor_ref"]) do
+          :ok -> Jason.encode!(%{accepted: true})
+          :error -> Jason.encode!(%{accepted: false})
+        end
```
brooksmtownsend, line 169: Backwards compatible 👍🏻

host_core/test/host_core/actors_test.exs
```diff
@@ -16,8 +16,75 @@ defmodule HostCore.ActorsTest do
...
+      HostCore.Actors.ActorSupervisor.terminate_actor(@echo_oci_key, 1)
+    end)
+
+    assert :error == HostCore.Actors.ActorSupervisor.live_update(@echo_oci_reference)
+  end
+
+  test "live update with new revision succeeds" do
+    {:ok, _pid} = HostCore.Actors.ActorSupervisor.start_actor_from_oci(@echo_old_oci_reference)
+
+    assert :ok == HostCore.Actors.ActorSupervisor.live_update(@echo_oci_reference)
```
brooksmtownsend, line 35: Thoughts on also including a scale test? e.g. start 5-10 instances of echo, live update, and then assert ok, check count, and ensure claims are of the updated revision?

brooksmtownsend (changes requested):
A couple changes and clarifying questions

host_core/lib/host_core/actors/actor_supervisor.ex
autodidaddict: this is synchronous code. The `with` statement basically lets you run a bunch of code in sequence, and if any of the pattern matches fail, then the `else` clause is run. [with](https://www.openmymind.net/Elixirs-With-Statement/)

host_core/test/host_core/actors_test.exs
autodidaddict: I'll be adding scale tests after we refactor the control interface to use scale operations rather than start/stop

brooksmtownsend: Sounds good to me 👍🏻

host_core/lib/host_core/actors/actor_supervisor.ex
brooksmtownsend: Thanks 🙂

brooksmtownsend (approved):

Issue #146, opened 2021-07-29
Update readiness and liveness probes to reflect status of NATS connections


autodidaddict:
Once #136 has been implemented, update the readiness and liveness probes to check the NATS connection and the status of the stream, ephemeral consumers, etc.

Issue #69, opened 2021-06-27
Support readiness and liveness probes


autodidaddict:
Support the ability to have `liveness` and `readiness` probes responses. Such probes are commonly used in automated scheduler environments like Kubernetes. 

If a liveness probe fails, it means that the process is not fully running and can be killed. Otherwise, the process is _up_ and to be considered _alive_ (can receive additional workloads like actors and providers).

If a readiness probe fails, it indicates that _at least one_ provider or actor within the host is reporting a status of unhealthy as of the last health check internal probe. This readiness probe will return HTTP 500 until all actors and providers in the host report healthy, either through "becoming" healthy or through eviction (the unhealthy entity is de-scheduled).

autodidaddict:
Vote that the following endpoints be used for plain HTTP access:
* `/healthz` - per industry convention, this is a liveness probe
* `/ready` - per semi-industry convention, this is a readiness probe

The `healthz` and `ready` endpoints are free to return whatever payload they wish, the only thing most schedulers (including Kubernetes) care about is the return of `HTTP 200` or `HTTP 500`.

autodidaddict:
For reference, [Spring Boot Actuator Endpoints](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html)

autodidaddict:
Per recent documentation from Kubernetes, they've retired the `healthz` probe endpoint and now only have `livez` and `readyz`, so we'll do the same by convention.

Also, more of a reminder to myself than anything else:
* Readiness probe failure indicates that the system is _temporarily_ in a bad/held/blocked state, and that anyone trying to use it should _try again_
* A liveness probe failure indicates that the system is in such a bad state that it should be _restarted_

PR #147, opened 2021-07-29
Readiness and liveness probes

autodidaddict:
I created #146 as a follow-on to #69 so that the probes will be updated once the application is using a durable JetStream, with the results of the probes reflecting the status of the RPC and control connections and the stream, etc.

### DIFF
--- a/host_core/lib/host_core/application.ex
+++ b/host_core/lib/host_core/application.ex
@@ -51,7 +51,14 @@ defmodule HostCore.Application do
       # {HostCore.Worker, arg}
       {Registry, keys: :unique, name: Registry.ProviderRegistry},
       {Registry, keys: :duplicate, name: Registry.ActorRegistry},
-      {HostCore.Host, config},
+      Supervisor.child_spec(
+        {Gnat.ConnectionSupervisor, HostCore.Nats.control_connection_settings(config)},
+        id: :control_connection_supervisor
+      ),
+      Supervisor.child_spec(
+        {Gnat.ConnectionSupervisor, HostCore.Nats.rpc_connection_settings(config)},
+        id: :rpc_connection_supervisor
+      ),
       {HostCore.HeartbeatEmitter, config},
       {HostCore.Providers.ProviderSupervisor, strategy: :one_for_one, name: ProviderRoot},
       {HostCore.Actors.ActorSupervisor, strategy: :one_for_one, name: ActorRoot},
@@ -119,7 +126,8 @@ defmodule HostCore.Application do
            ]
          }},
         id: :latticectl_consumer_supervisor
-      )
+      ),
+      {HostCore.Host, config}
     ]
 
     # See https://hexdocs.pm/elixir/Supervisor.html
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -35,7 +35,7 @@ defmodule HostCore.Host do
   """
   @impl true
   def init(opts) do
-    start_gnat(opts)
+    # start_gnat(opts)
     configure_ets()
 
     :ets.insert(:config_table, {:config, opts})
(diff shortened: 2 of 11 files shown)


autodidaddict:
Testing:

```
❯ curl localhost:4000/api/livez
{"live":true}
~ 
❯ curl localhost:4000/api/readyz
{"ready":true}
~ 
```

host_core/lib/host_core/host.ex
```diff
@@ -35,7 +35,7 @@ defmodule HostCore.Host do
   """
   @impl true
   def init(opts) do
-    start_gnat(opts)
+    # start_gnat(opts)
```
brooksmtownsend, line 38: we can remove this now, right?

host_core/lib/host_core/application.ex
```diff
@@ -119,7 +126,8 @@ defmodule HostCore.Application do
            ]
          }},
         id: :latticectl_consumer_supervisor
-      )
+      ),
+      {HostCore.Host, config}
```
brooksmtownsend, line 130: Were you able to do any testing as to if this affects the ability for the host to query cache, or does this effectively break the cache query entirely?

brooksmtownsend (changes requested):

host_core/lib/host_core/application.ex
autodidaddict: I'm reasonably convinced this is going to break the cache. I can see the host answering this query. This problem will go away when we switch to the JetStream-based cache since we won't be asking hosts to answer the query, we'll be subscribing to the ephemeral consumers.

host_core/lib/host_core/host.ex
autodidaddict: already removed ;)

host_core/lib/host_core/host.ex
```diff
@@ -52,6 +51,7 @@ defmodule HostCore.Host do
 
   @impl true
   def handle_continue(:query_lattice_cache, state) do
+    Process.sleep(300)
```
brooksmtownsend, line 54: This will likely guarantee that a host can request itself for the lattice cache, feels wrong to me. I think you can remove this `Process.sleep` without changing the fact that we are not guaranteeing proper cache retrieval. Not sure how far jetstream is away from us now, but I don't like that this leaves proper cache retrieval up to getting lucky on the queue subscribe

autodidaddict: So the core part of the problem here is that there's a race condition, without this sleep it's entirely possible to attempt to request the cache contents before the connection supervisor(s) have prepared the NATS connection(s). Using the connection supervisor versus a regular connection apparently introduces some delay in when the connection starts.

autodidaddict: Without this, I can get the tests to fail locally on my machine about 30% of the time just by attempting to query the cache before the NATS connections are available.

autodidaddict: we should, in theory, have a version of `Gnat` that talks to JetStream in a week or so.

brooksmtownsend: Ah so the issue is that we can't actually _request_ the claims because the nats connection isn't ready yet, but if we wait for the nats connection then we might actually have already subscribed to the linkdef.get. 

Stupid computers.

host_core/test/host_core/actors_test.exs
```diff
@@ -215,6 +215,10 @@ defmodule HostCore.ActorsTest do
   end
 
   test "can invoke via call alias" do
+    # TODO
+    # there's no way we should need this.
+    Process.sleep(500)
```
brooksmtownsend, line 220: Can we remove this and everything works still?

brooksmtownsend (changes requested):
only one request to remove a hopefully unnecessary timer

host_core/test/host_core/actors_test.exs
autodidaddict: Good catch

brooksmtownsend (approved):

PR #150, opened 2021-07-30
Enforce provider archive

brooksmtownsend:
Fixes #150 

This PR makes the `start_executable_provider` method private and enforces that publicly users of `host_core` or `wasmcloud_host` must start providers that are inside of a provider archive file. This is to ensure that all providers that are started in a wasmcloud runtime are signed with embedded claims.

Now, in the wasmcloud dashboard, the only possible files to upload as a file are `.par` files and `.gz` files, and they are loaded in as provider archives. Failures to submit proper files should be reported in the UI.


autodidaddict:
Not sure .. looks like tests are failing?

autodidaddict:
@brooksmtownsend can we get the tests passing?

brooksmtownsend:
@autodidaddict ready for review now. The current provider start behavior now is:
1. Get provider archive bytes, either from OCI or from file
2. convert bytes into provider archive wrapper object
3. extract executable to cache directory
    - If the file does not exist or if the provider is not running, write executable to cache
    - Otherwise (file exists and provider is running) simply return the cache directory

autodidaddict (approved):

Issue #53, opened 2021-06-17
Host must handle process termination from managed providers


autodidaddict:
The host runtime needs to monitor the process (`Port`) created when a managed capability provider is executed. If that process terminates, the host needs to react to this by "removing" the provider from anything that would indicate the provider is up and running in the runtime.

It should also publish the "provider terminated" event.

In any event, if a provider process terminates before the host asks it to do so, then this is an unscheduled termination and should be flagged as an error condition.

PR #151, opened 2021-08-03
Enabling port monitoring of provider executable

autodidaddict:

### DIFF
--- a/host_core/lib/host_core/providers/provider_module.ex
+++ b/host_core/lib/host_core/providers/provider_module.ex
@@ -56,6 +56,7 @@ defmodule HostCore.Providers.ProviderModule do
       |> to_charlist()
 
     port = Port.open({:spawn, "#{path}"}, [:binary])
+    Port.monitor(port)
     Port.command(port, "#{host_info}\n")
 
     {:os_pid, pid} = Port.info(port, :os_pid)
@@ -92,7 +93,7 @@ defmodule HostCore.Providers.ProviderModule do
       System.cmd("kill", ["-9", "#{state.os_pid}"])
     end
 
-    publish_provider_stopped(state.public_key, state.link_name, state.instance_id)
+    publish_provider_stopped(state.public_key, state.link_name, state.instance_id, "normal")
     {:stop, :normal, :ok, state}
   end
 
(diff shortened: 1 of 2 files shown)


autodidaddict:
Closes #53

host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -113,6 +114,18 @@ defmodule HostCore.Providers.ProviderModule do
...
+
+    {:noreply, state}
+  end
+
+  def handle_info({:DOWN, _ref, :port, _port, reason}, state) do
+    Logger.error("Received DOWN message from port (executable stopped) - #{reason}")
+    publish_provider_stopped(state.public_key, state.link_name, state.instance_id, "#{reason}")
+    {:stop, reason, state}
+  end
+
```
brooksmtownsend, line 128: Is there a reason to not publish that the provider stopped even if it stopped normally?

host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -113,6 +114,18 @@ defmodule HostCore.Providers.ProviderModule do
...
+    Logger.debug("Received DOWN message from port (executable stopped normally)")
+
+    {:noreply, state}
+  end
+
+  def handle_info({:DOWN, _ref, :port, _port, reason}, state) do
+    Logger.error("Received DOWN message from port (executable stopped) - #{reason}")
+    publish_provider_stopped(state.public_key, state.link_name, state.instance_id, "#{reason}")
+    {:stop, reason, state}
+  end
```
brooksmtownsend, line 127: If a provider fails to start in its startup routine (e.g. if we start the provider that fails to deserialize the base64 encoded HostData) would that be handled by this handler? If so, that's great.

host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -185,14 +198,16 @@ defmodule HostCore.Providers.ProviderModule do
     Gnat.pub(:control_nats, topic, msg)
   end
 
-  def publish_provider_stopped(public_key, link_name, instance_id) do
+  @spec publish_provider_stopped(String.t(), String.t(), String.t(), String.t()) :: :ok
+  def publish_provider_stopped(public_key, link_name, instance_id, reason) do
```
brooksmtownsend, line 202: Nice to see these `@spec`s starting to work their way in

brooksmtownsend (commented):
A couple questions but no changes specifically

host_core/lib/host_core/providers/provider_module.ex
autodidaddict: Because if the provider stops normally, then there's another function that's been called that will publish the stop. We're not publishing the stop here because that would actually result in a double-publish of the event.

host_core/lib/host_core/providers/provider_module.ex
autodidaddict: yeah, if the provider crashes reading stdin, this will be triggered.

host_core/lib/host_core/providers/provider_module.ex
brooksmtownsend: Ah ok, so stopping with a reason of `:normal` means we essentially told that port to go DOWN, any other reason (like being killed by the OS, failing to startup) falls under the other handler. 👍🏻

autodidaddict: In other words, the only reason we'd ever get the `:normal` shutdown from the child port is because the parent process (the provider module) was told to terminate normally, in which case it publishes the provider stopped event.

brooksmtownsend (approved):

PR #152, opened 2021-08-03
Refactored control interface calls to respond appropriately

brooksmtownsend:
In a previous refactor, `Gnat.pub` calls were converted to implicit replies in the message handers for the control interface. This PR fixes an issue where some handlers would simply return `:ok` instead of appropriately responding to the control interface queries.

### DIFF
--- a/host_core/lib/host_core/control_interface/acl.ex
+++ b/host_core/lib/host_core/control_interface/acl.ex
@@ -1,23 +1,30 @@
 defmodule HostCore.ControlInterface.ACL do
   def all_actors() do
     HostCore.Actors.ActorSupervisor.all_actors()
-    |> Enum.map(fn {k, v} ->
-      %{
-        id: k,
-        revision: get_revision(k),
-        image_ref: get_image_ref(k),
-        instance_id: HostCore.Actors.ActorModule.instance_id(v)
-      }
+    |> Enum.flat_map(fn {id, pids} ->
+      revision = get_revision(id)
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/control_interface/acl.ex
```diff
@@ -1,23 +1,30 @@
...
 
   def all_providers() do
+    # TODO: retrieve revision information for provider
     HostCore.Providers.ProviderSupervisor.all_providers()
     |> Enum.map(fn {pk, link, _contract, instance_id} ->
       %{
         id: pk,
         link_name: link,
-        revision: get_revision(pk),
+        revision: 0,
```
brooksmtownsend, line 27: I left this with a TODO as the claims information for providers is not being stored in the `:claims_table`. There's a couple different ways I could see solving for this, looking to get some thoughts:

1. store provider claims in the claims table under the provider_key as a key (different claims structure)
2. store the revision information as ProviderModule state
3. store the revision information in the `Registry.ProviderRegistry` along with instance id, pub key, etc.

I'm leaning towards one of these but curious if there are other thoughts.

autodidaddict: Option 1 is where I'm leaning (we should create a story to do this at some point). Option 3 is non-idiomatic as those kinds of registries are really designed to store pids.

autodidaddict (approved):

brooksmtownsend:
Documented provider claims issue in #153

PR #156, opened 2021-08-05
Allow deleting providers, scaling actors from dashboard

### LINKED ISSUE #85: Replace Start/Stop actor messages in control interface with a single "Scale Actor" message
The `StartActor` and `StopActor` messages are potentially a bit outdated now that we support multiple actors of the same public key within a single host (OTP). It might be a better control interface to expose a single scale message that contains the desired instance count and if the actor isn't currently running, then we start that actor for the first time if the scale count is > 0.

### LINKED ISSUE #132: Implement stopping actors and providers in washboard

brooksmtownsend:
Fixes #132, fixes #85 

This PR allows deleting providers from the UI, and scaling actors up and down from the UI. The UI will now handle actor/provider stopped events appropriately, so it will update even if the stop command doesn't originate from the dashboard.

I've also implemented the control interface handler for scale actor, which is an addition but doesn't break anything about the current control interface. Its topic is `wasmbus.ctl.#{lattice_prefix}.cmd.#{host_id}.scale` and expects an actor ID, actor OCI reference, and desired number of replicas. I've designed the scale actor command to work regardless of ocireference for no scaling and scaling down, but in order to scale up an ocireference must be supplied.

If you attempt to scale an actor up that was started from a file in the UI, you will be shown an error message saying that's not supported at this time.

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -206,8 +206,7 @@ defmodule HostCore.Actors.ActorModule do
 
     Logger.info("Subscribing to #{topic}")
     {:ok, subscription} = Gnat.sub(:lattice_nats, self(), topic, queue_group: topic)
-    Agent.update(agent, fn state -> %State{state | subscription: subscription} end)
-    Agent.update(agent, fn state -> %State{state | ociref: oci} end)
+    Agent.update(agent, fn state -> %State{state | subscription: subscription, ociref: oci} end)
 
     Process.send_after(self(), :do_health, @thirty_seconds)
 
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -21,7 +21,8 @@ defmodule HostCore.Actors.ActorModule do
       :api_version,
       :invocation,
       :claims,
-      :subscription
+      :subscription,
+      :ociref
     ]
   end
 
@@ -50,6 +51,10 @@ defmodule HostCore.Actors.ActorModule do
 
   def instance_id(pid) do
     GenServer.call(pid, :get_instance_id)
+  end
+
+  def ociref(pid) do
+    GenServer.call(pid, :get_ociref)
   end
 
   def halt(pid) do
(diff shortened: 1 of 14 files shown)


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -198,6 +207,7 @@ defmodule HostCore.Actors.ActorModule do
     Logger.info("Subscribing to #{topic}")
     {:ok, subscription} = Gnat.sub(:lattice_nats, self(), topic, queue_group: topic)
     Agent.update(agent, fn state -> %State{state | subscription: subscription} end)
+    Agent.update(agent, fn state -> %State{state | ociref: oci} end)
```
autodidaddict, line 210: You can combine these 2 `Agent.update` calls into one. It's a tiny performance benefit but I think it'll look cleaner.

host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -92,6 +92,59 @@ defmodule HostCore.Actors.ActorSupervisor do
...
+        true ->
+          ""
+      end
+
+    diff = current_count - desired_count
+
+    cond do
+      # Current count is desired actor count
+      diff == 0 ->
+        {:ok}
```
autodidaddict, line 121: idiomatic here would be for the type of this fucntion to be `:ok | {:error, term}` rather than `{:ok} | {...}`  It's fine for the ok not to be in a tuple if there's no return value.

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -121,6 +121,43 @@ defmodule HostCore.ControlInterface.Server do
...
+  defp handle_request({"cmd", host_id, "scale"}, body, _reply_to) do
+    scale_request = Jason.decode!(body)
+
+    if host_id == HostCore.Host.host_key() do
+      actor_id = scale_request["actor_id"]
+      actor_ref = scale_request["actor_ref"]
+      replicas = String.to_integer(scale_request["replicas"])
+
+      case HostCore.Actors.ActorSupervisor.scale_actor(actor_id, replicas, actor_ref) do
+        {:ok} ->
```
autodidaddict, line 135: see previous comment, we can take the tuple wrapper off of this `:ok`

wasmcloud_host/assets/vendor/wasmcloud/js/tooltips.js
```diff
@@ -0,0 +1,19 @@
+import { Tooltip } from "../../@coreui/coreui-pro/js/coreui.bundle.min";
```
autodidaddict, line 1: What horrible language is this??

autodidaddict (changes requested):
Just 2 tiny recommendations otherwise LGTM

autodidaddict (approved):

PR #159, opened 2021-08-06
Support OCI environment variables

brooksmtownsend:
This PR introduces the `WASMCLOUD_OCI_ALLOW_LATEST` and `WASMCLOUD_OCI_ALLOWED_INSECURE` environment variables. allow_latest is a boolean to allow the `:latest` tag, and allowed_insecure is a comma-separated list of registries that should be accessed over HTTP rather than HTTPS. 

I also changed the `par_from_bytes` function to return `{:ok, par}` rather than just `par` so we can actually match against the `ok` condition.

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -32,7 +32,11 @@ defmodule HostCore.Actors.ActorSupervisor do
   end
 
   def start_actor_from_oci(oci) do
-    case HostCore.WasmCloud.Native.get_oci_bytes(oci, HostCore.Oci.allow_latest(), HostCore.Oci.allowed_insecure()) do
+    case HostCore.WasmCloud.Native.get_oci_bytes(
+           oci,
+           HostCore.Oci.allow_latest(),
+           HostCore.Oci.allowed_insecure()
+         ) do
       {:error, err} ->
         Logger.error("Failed to download OCI bytes for #{oci}")
         {:stop, err}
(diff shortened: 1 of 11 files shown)


stevelr:
tested without the env variable and load from insecure oci failed (as expected)
tested with env variable and load from insecure oci succeeded
also tested env variable specifity: adding allowed-insecure "localhost:5001" but trying to pull from "localhost:5000" insecure - failed as expected

I don't know if this is something that should be fixed or not: If a previous run of the host loaded an actor from an insecure oci server, and then the host is restarted without the allowed-insecure environment variable, the start actor command will succeed with the insecure oci url, because the actor is pulled from the cache instead of the url.

brooksmtownsend:
>I don't know if this is something that should be fixed or not: If a previous run of the host loaded an actor from an insecure oci server, and then the host is restarted without the allowed-insecure environment variable, the start actor command will succeed with the insecure oci url, because the actor is pulled from the cache instead of the url.

This is an interesting edge case. Not sure if we have to fix it now, if we did I don't think having an extra branch of code for "if insecure and exists in cache" is the right idea, this feels more like something that we should address by clearing the cache on host shutdown or something of the sort. Perhaps this will be best filed as an issue and labeled with `question` @stevelr? (I'm mostly proposing that because I don't have a solution or a strong feeling on how to address it)

brooksmtownsend:
Thanks for doing some integration testing 😄

stevelr:
In general, we shouldn't be using environment variables for settings that may change during the (hopefully long) runtime of the host, unless there is also a way to change the setting at runtime. In this case, however, I like the inflexibility because it's more secure. A production environment should be disabling loading from insecure oci connections (as is the default), and preventing a runtime reconfiguration that allows loading from insecure oci connections is a good thing.

stevelr (approved):
based on preliminary testing, looks good.

PR #163, opened 2021-08-10
Adding link definitions to HOST_DATA

autodidaddict:
@brooksmtownsend I couldn't get the cross builds to work, any chance you could regenerate the PARjeezies in this branch? (I also lack the private keys to produce the "right" ones).

@stevelr see the definition for the `LinkDefinition` structs in the Rust and Go example providers for the details you might need to support the changes to the host data struct in any generated code.


brooksmtownsend:
@autodidaddict working on rebuilding the pars now, I think I found out why your cross build won't work. They still haven't accepted Chris's PR for the apple darwin target, here's a command you can use:
```
cargo install --git https://github.com/ChrisRx/cross --branch add-darwin-target --force
```

brooksmtownsend (approved):
LGTM, recompiled parJEEzies

PR #170, opened 2021-08-12
Adding configurable RPC timeouts

autodidaddict:
Worth noting that we have not added a timeout to the `linkdefs.put` and `linkdefs.del` RPC calls because those are publishes, not requests, and that's because it's entirely possible (and likely) that capability providers may not actually be up and running at the time the definition is removed or created.

host call RPC timeouts, however, are applicable.

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -115,7 +115,7 @@ defmodule HostCore.Host do
   end
 
   def rpc_timeout() do
-    case :ets.lookup(:rpc_timeout, :config) do
+    case :ets.lookup(:config_table, :config) do
       [config: config_map] -> config_map[:rpc_timeout]
       _ -> 2_000
     end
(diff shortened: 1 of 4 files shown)


brooksmtownsend (dismissed):
LGTM, I'm a fan on the `_MS` prefix too to avoid ambiguity

Edit: Looks like tests are failing on the config lookup

autodidaddict:
Woops, this looks like a typo I fixed locally that didn't get checked in.

host_core/lib/host_core/host.ex
```diff
@@ -114,6 +114,13 @@ defmodule HostCore.Host do
     end
   end
 
+  def rpc_timeout() do
+    case :ets.lookup(:config_table, :config) do
```
brooksmtownsend, line 118: I've actually gotten burned by looking up the key in the map too. 🔥

brooksmtownsend (approved):

Issue #174, opened 2021-08-14
bugs in random numbergen


stevelr:
In numbergen.ex:
```
def invoke("RandomInRange", payload) do
    params = Msgpax.unpack!(payload)

    min = min(params["min"], 0)
    max = max(params["max"], 4_294_967_295)
    Msgpax.pack!(Enum.random(min..max))
end
```

min() should be max(), and max() should be min()
and the upper bound is defined (in the smithy interface) to be exclusive, but Enum.random range is inclusive, so, I think the implementation should be like this:
```
def invoke("RandomInRange", payload) do
    params = Msgpax.unpack!(payload)

    min = max(params["min"], 0)
    max = min(params["max"]-1, 4_294_967_295)
    Msgpax.pack!(Enum.random(min..max))
end
```
Also, bug(s) here:

```
 def invoke("Random32", _payload) do
    Msgpax.pack!(floor(:rand.uniform(4_294_967_294)))
  end
```

rand.uniform returns a number from 1<=n<=max, and it's an int so I don't know why you need floor().  I think this function should be

```
 def invoke("Random32", _payload) do
    Msgpax.pack!(:rand.uniform(4_294_967_296)-1)
  end
```
or `Enum.random(0..4_294_967_295)` but I don't know the tradeoffs between Enum.random and :rand.uniform.

autodidaddict:
Woopsy on the use of `floor`, I must've been reading the docs for `uniform/0`, which returns a float, whereas `uniform/1` returns an int.

PR #175, opened 2021-08-14
Closing #174 - fixing bugs in numbergen

autodidaddict:


host_core/lib/host_core/providers/builtin/numbergen.ex
```diff
@@ -1,17 +1,17 @@
...
 
   def invoke("RandomInRange", payload) do
     params = Msgpax.unpack!(payload)
 
-    min = min(params["min"], 0)
-    max = max(params["max"], 4_294_967_295)
-    Msgpax.pack!(Enum.random(min..max))
+    min = max(params["min"], 0)
+    max = min(params["max"] - 1, 4_294_967_295)
+    IO.iodata_to_binary(Msgpax.pack!(Enum.random(min..max)))
```
brooksmtownsend, line 11: So our random generator is in the range of `[0..MAX)`, and can't be the max number? The only thing that seems weird to me here is that if I supply `4_294_967_295` as the max number, then the highest number it will generate is one less than that (`4_294_967_294`), but if I supply a bigger number like `4_999_999_999` it will generate up to the max U32

I'd suggest removing the subtracting 1, since `Enum.random/1` is inclusive on both ends

brooksmtownsend (changes requested):

brooksmtownsend (approved):
LGTM

Issue #173, opened 2021-08-14
no log output when actor invokes logging provider


stevelr:
This is my first attempt to use the builtin logging provider, so it could be a bug in the actor or the otp host.
When I invoke WriteLog, I see debug messages from the host:

```
[debug] Guest request
[debug] Got current invocation
[debug] host call: wasmcloud:builtin:logging - default: WriteLog (25 bytes)
[debug] host call: wasmcloud:builtin:logging - default: WriteLog (28 bytes)
[debug] host call: wasmcloud:builtin:logging - default: WriteLog (32 bytes)
[debug] host call: wasmcloud:builtin:logging - default: WriteLog (58 bytes)
```
The actor wasm is signed with the capability "wasmcloud:builtin:logging", which I confirmed with `wash claims inspect`.

I tried invoking Linkdefs.Manager.put_link_definition to bind the actor to the logging interface, but that produces the error: 

```
Invalid provider spec: no provider with contract wasmcloud:builtin:logging
```

Is it possible that the host isn't performing the linking to builtin providers when the actor is started, or that the host doesn't know that the builtin providers are available?

other details:
WriteLog invokes __host_call, as it does for other providers, with "wasmcloud:builtin:logger" and "default".

The message is a messagepack structure containing "level" and "text", both strings, which looks like the structure that is being used in logger.ex.

stevelr:
Here's additional information. I wasn't logging the error returned:
```
 Host send error Invocation not authorized
 ```

I just tried an actor with the numbergen builtin provider and got back the same error, even though the actor is signed with the capability.

autodidaddict:
Builtins don't require link defs in otp. Will investigate the rest.

autodidaddict:
Your own comment says you're invoking wasmcloud:builtin:logger but the contract I'd is logging,  not logger. Was that a typo in the issue?

autodidaddict:
Looks like this is coming from logic that denies authorization to all invocations that don't have a corresponding linkdef, and builtins don't require linkdefs in the OTP version. Will have a PR to fix this shortly.

stevelr:
tested on "bugs/nobuiltins" branch

PR #176, opened 2021-08-14
Closing #173 - authorization denied for builtin providers

autodidaddict:


stevelr:
moved verify_link above the target `:unknown` check, because builtins have target unknown.
had to add authorize_link for builtins also.
These changes made both logging and numbergen calls from actors work.
Thanks for your fix this morning - it pointed me in the right direction.

stevelr:
commit 4dc3d4d allowed the actor test cases to work, but introduced a new bug that the actor is authorized to use a builtin provider even if its wasm wasn't signed with that capability - looking into this

stevelr:
commit f912fe5 fixes authorization for builtins that are in claims

stevelr (approved):
These changes partially fixed the problem and pointed me in the right direction. I added a couple more commits to this branch to address remaining issues.

brooksmtownsend (approved):
Clean ✅

autodidaddict:
@stevelr changes made after the initial pull request have broken the tests. It looks like the test currently failing is invocation by call alias.

autodidaddict:
That used to be there.

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -227,6 +227,11 @@ defmodule HostCore.WebAssembly.Imports do
     %{token | authorized: Enum.member?(claims.caps, namespace)}
   end
 
+  # allow actor-to-actor calls
+  defp authorize_call(token = %{target: {:actor, _, _}}) do
+    %{token | authorized: true}
+  end
```
brooksmtownsend, line 233: So if the target is an actor, claims and namespace are irrelevant. Looks like this is what we had before 👍🏻

brooksmtownsend (approved):

host_core/lib/host_core/web_assembly/imports.ex
autodidaddict: Yep, at least until we support extensibility with OPA calls

PR #189, opened 2021-08-18
Adding support for .env hierarchy of files to supply environment configuration

autodidaddict:
This is not the ultimate long-term goal. In the long term I'd like to have an actual configuration file that _properly_ merges with supplied environment variables. However, due to some limitations in the existing Elixir libraries, I'm going to pass on implementing the configuration file and, to get us to MVP, add the support for `.env` , `.env.local`, `.env.{ENV}`, and `.env.{ENV}.local` files.


brooksmtownsend (approved):
So, just to confirm, the behavior now is that the environment variables and the .env file will smart merge, and the .env file will take precedence if variables are declared twice?

brooksmtownsend:
The PR looks good so I approved, we just went through a few iterations on slack so I wanted to confirm the behavior

autodidaddict:
The regular environment variables will override anything in a .env file. This is basic standard design for .env files.

PR #190, opened 2021-08-19
Using contract ID and link name for provider executable path

autodidaddict:
See the linked issue for discussion around why I ended up going with the contract Id and link name for the filename instead of instance IDs.


brooksmtownsend (approved):
LGTM

PR #192, opened 2021-08-19
Lattice dashboard view

### LINKED ISSUE #68: Implement "Lattice" view in Dashboard
Currently, the dashboard shows an overview of a single `wasmcloud_host`, though you cannot access multiple hosts in the same dashboard. After #73 , multiple hosts will be able to run on different ports on the same machine, but as it is today multiple hosts can be connected to the same lattice.

If it's technically possible, the wasmcloud dashboard should be able to display a "lattice" view that shows all hosts connected to the same lattice, allowing remote control of each of them. I can see this being technically challenging until #5 is implemented, so it may depend on that issue before we can issue commands to remote hosts via the dashboard.

brooksmtownsend:
Fixes #68 

Most of the changes in this PR revolve around moving the metrics into a separate tab and adding logic to select a host from the list of discovered hosts in the UI. When you select a host, the `Host Info` card will appear with that hosts' labels, lattice prefix, a boolean denoting if it's the "local" host (e.g. the host launched with the dashboard) and a total count of actors and providers running on that host.

Looking for feedback on the code as well as how the UI looks. Does the host selector warrant a different color for the "local" host? Is the UI getting too crowded? Etc.

### DIFF
--- a/wasmcloud_host/assets/css/app.scss
+++ b/wasmcloud_host/assets/css/app.scss
@@ -1,6 +1,7 @@
 /* This file is for your main application css. */
 @import "./coreui/style.min.css";
 
+
 /* LiveView specific classes for your customizations */
 .phx-no-feedback.invalid-feedback,
 .phx-no-feedback .invalid-feedback {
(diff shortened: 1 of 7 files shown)


brooksmtownsend:
Some screenshots:
All hosts shown
image

One host selected (host info card appears)
image

Metrics tab
image

autodidaddict (approved):

PR #199, opened 2021-08-27
Make tests deterministic

### LINKED ISSUE #121: make tests deterministic without calls to sleep()
Several of the wasmcloud tests have sleep statements to wait for the host to reach a particular state. Setting the amount of time to sleep is a trial-and-error process and either not reliable or causes the tests to take longer than necessary. There should be a way to get events from the server to know when it has reached a particular state. 

Fixing this PR requires the server to emit enough events so that tests can detect when a state has been reached. Keep in mind security implications of the events; a test key may be required to allow tests to have access to the applicable events without exposing them to others in the lattice.

The goal is minimize the impact on test authors so that it's still easy to write tests. It should take one line of code, or close to it, to wait for an event. This probably needs a set of library functions that test authors can use: `wait_for_startup()`, `wait_for_actor_load()`, etc.  - where internally, the function subscribes to the applicable event channel with the required event filter, and waits, returning to caller.

brooksmtownsend:
Fixes #121 

This PR introduces the `EventWatcher` support module, which is a `GenServer` that:
1. Registers interest in `cache_loader` events to receive linkdefs, ocirefs, and claims
2. Listens for wasmbus events and stores them in an in-memory list

This `EventWatcher` can easily be instantiated as a part of the setup piece of each test, and can be used to assert that certain actions have taken place. The `EventWatcher` can:
1. Assert whether or not a certain event has been received, e.g. `ActorStarted`
2. Find all matching events with a given data parameter, e.g. `%{ actor_id: "MASDASD"}`
3. Wait for an event to be received that matches a given (optional) data map. This can be used to wait for any event that is published to the `wasmbus.evt.#{prefix}` topic

In order to make this a little easier for test writers, I've created a few common helper functions
1. `wait_for_actor_start` which only requires the `EventWatcher` pid, and the actor public key
2. `wait_for_actor_stop`, same as above
3. `wait_for_provider_start` requires pid, contract_id, link_name, and public_key
4. `wait_for_provider_stop` required pid, link_name, public_key
5. `wait_for_linkdef` requires pid, actor_id, contract_id, link_name

All of the above timeout after 30 seconds, and retry every second until the timeout is reached. A sample usage of this is as follows:

```
    {:ok, bytes} = File.read("test/fixtures/actors/echo_s.wasm")
    {:ok, _pid} = HostCore.Actors.ActorSupervisor.start_actor(bytes)
    on_exit(fn -> HostCore.Actors.ActorSupervisor.terminate_actor(@echo_key, 1) end)

    :ok = HostCoreTest.EventWatcher.wait_for_actor_start(evt_watcher, @echo_key)
```

### Note
This PR sat for a while as I debugged some NATS Jetstream issue, and my best hypothesis as to what's going wrong is that there's some kind of race condition where the event watcher can initialize quicker than the host can, and that causes issues for the test. I've fixed this by adding a sleep for 2 seconds to allow for the first ping/pong interaction during event watcher initialization, and after 5+ consecutive test suite runs I can't reproduce the problem anymore.

### DIFF
--- a/.github/workflows/build_tarballs_manual.yml
+++ b/.github/workflows/build_tarballs_manual.yml
@@ -146,7 +146,6 @@ jobs:
           push: false # Note here that the image will NOT be pushed to our container registries
           load: true
           file: ${{env.working-directory}}/Dockerfile.release
-          outputs: type=docker,dest=/tmp/wasmcloud_host.tar
           build-args: |
             BUILD_IMAGE=wasmcloud.azurecr.io/build/wasmcloud_host_tarball:${{ env.wasmcloud_host_version }}-x86_64-linux-gnu
             RELEASE_IMAGE=debian:buster-slim 
@@ -171,12 +170,6 @@ jobs:
         with:
           name: x86_64-linux-gnu
           path: ${{env.working-directory}}/rel/artifacts/wasmcloud_host/${{ env.wasmcloud_host_version }}/x86_64-linux-gnu.tar.gz
-
-      - name: Upload docker image
-        uses: actions/upload-artifact@v2
-        with:
-          name: wasmcloud_host-v${{ env.wasmcloud_host_version }}-docker
-          path: /tmp/wasmcloud_host.tar
 
   release-macos:
     name: Release MacOS Tarball
--- a/.github/workflows/build_tarballs_manual.yml
+++ b/.github/workflows/build_tarballs_manual.yml
@@ -146,6 +146,7 @@ jobs:
           push: false # Note here that the image will NOT be pushed to our container registries
           load: true
           file: ${{env.working-directory}}/Dockerfile.release
+          outputs: type=docker,dest=/tmp/wasmcloud_host.tar
           build-args: |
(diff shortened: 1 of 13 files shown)


brooksmtownsend:
I feel confident about the actor/provider event logic, what's causing intermittent failures here is the fact that linkdefs, claims, ocimaps persist across tests so they can't always be reliably asserted. Considering a few solutions here, knee jerk idea is to clear the cache between test runs, but I would want to replicate a real environment so that cache clear should happen in jetstream, not just in the event watcher

autodidaddict:
There is a special nats topic you can post a message to in order to purge a stream, you just need to know the lattice prefix to generate the stream name.

autodidaddict:
Not sure what's up, but one of the e2e tests failed in one of the scenarios:
```

  3) test echo roundtrip (HostCore.E2E.EchoTest)
Error:      test/e2e/echo_test.exs:17
     ** (MatchError) no match of right hand side value: {:error, :timeout}
     code: :ok =
     stacktrace:
       test/e2e/echo_test.exs:65: (test)
```

autodidaddict:
Did we ever track down the ping/disconnect issue with JetStream?

autodidaddict (approved):

PR #204, opened 2021-08-31
accept new-style method prefixes and old-style method-only names.

stevelr:
accept both "Service.Method" and "Method" for invoking methods, e.g., "NumberGen.Random32" and "Random32"

This change allows the host to support new code and is backwards-compatible with old (yesterday's) actors. Once all the interfaces have been regenerated and all the examples updated, we can remove the support for non-prefixed method names.

Signed-off-by: stevelr <steve@cosmonic.com>

### DIFF
--- a/host_core/lib/host_core/providers/builtin/logging.ex
+++ b/host_core/lib/host_core/providers/builtin/logging.ex
@@ -1,7 +1,7 @@
 defmodule HostCore.Providers.Builtin.Logging do
   require Logger
 
-  def invoke(actor, method, payload) when method == "Logging.WriteLog" or method == "WriteLog" do
+  def invoke(actor, method, payload) when method in ["Logging.WriteLog", "WriteLog"] do
     msg = Msgpax.unpack!(payload)
     text = "[#{actor}] #{msg["text"]}"
 
--- a/host_core/lib/host_core/providers/builtin/logging.ex
+++ b/host_core/lib/host_core/providers/builtin/logging.ex
@@ -1,7 +1,7 @@
 defmodule HostCore.Providers.Builtin.Logging do
   require Logger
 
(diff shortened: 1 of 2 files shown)


autodidaddict:
Can't tell from mobile- why did a ton of the tests fail?

autodidaddict:
Ha looks like you might be missing a mix format call

stevelr:
haha - I edited those in vim and haven't installed any vim plugins for elixir yet! I did confirm they passed the builtin provider tests though :-)

autodidaddict:
I'm sorry but when running formatters one is required to make a pun or humorous equivalent e.g. "this commit is for Matt"

stevelr:
Matt's cat Bergen had to have surgery yesterday and the pain meds didn't wear off for three days.  I was supposed to be Best Man for Matt's wedding, but the whole thing was cancelled because of numb Bergen.

autodidaddict:
I should report this to github

host_core/lib/host_core/providers/builtin/logging.ex
```diff
@@ -1,7 +1,7 @@
 defmodule HostCore.Providers.Builtin.Logging do
   require Logger
 
-  def invoke(actor, "WriteLog", payload) do
+  def invoke(actor, method, payload) when method == "Logging.WriteLog" or method == "WriteLog" do
```
autodidaddict, line 4: I actually think you can rewrite this to:

`def invoke(actor, method, payload) when method in ["Logging.WriteLog", "WriteLog"] do`

autodidaddict (commented):
LGTM, you can refactor to use the `in` operator in the guard clauses if you like.

autodidaddict (approved):

PR #214, opened 2021-09-14
Update test fixtures

### LINKED ISSUE #213: Upgrade test fixtures to newest providers
In order to verify that our own test suite is testing the same things that our developers will be experiencing as they go through tutorials and documentation, we need to upgrade the fixtures in our test suite.

* Upgrade the OCI reference of HTTP server to `0.14.x`
* Upgrade the fixture version of `httpserver.par.gz` to `0.14.x`
* Upgrade the figure version of `echo_s.wasm`, `kvcounter_s.wasm`, `kvcounter_unpriv_s.wasm` to all return the right http response that is compatible with http server `0.14.x`
* Upgrade the OCI reference to the aforementioned actors to whatever the new version is that is compatible with httpserver `0.14.x`

brooksmtownsend:
Fixes #213 

This PR updates test fixtures to actors and providers created for the `0.50` release of wasmcloud. This better establishes functionality through unit tests that represent our first party resources.

This PR also abstracts all constants in the test code to the module `HostCoreTest.Constants`, which is accessible from all tests. This allows us to change resources and variables in that one module, and it will update for all of the tests. Constants are still used in individual tests to create better readability.

To keep us honest, I've left some of the tests (notably `retrieves provider archive from OCI image`) with some hardcoded values such as `assert par.claims.version == "0.13.0"`. I could certainly abstract that into the Constants module, but for now it ensures we update tests appropriately. (further down in the file we assert byte size of the executables, this will likely change every time we update the `httpserver` provider archive anyways.)

I've also gone through and removed the various `on_exit` calls across test modules, replacing them with a single request to purge the host on test exit. This removes any edge cases where we get a provider ID slightly wrong, issue the wrong number of terminate requests, etc.


brooksmtownsend:
Tests are failing at the moment because the `kvcounter_unpriv` module needs to be a module that has the HTTPServer capability and attempts to invoke a capability afterwards, I'll have to custom build an actor for this test (will be essentially a clone of the kvcounter example, I'll make a custom makefile to make it easy to build)

host_core/test/e2e/controlinterface_test.exs
```diff
@@ -1,15 +1,14 @@
...
-  @redis_contract "wasmcloud:keyvalue"
+  @echo_key HostCoreTest.Constants.echo_key()
+  @kvcounter_key HostCoreTest.Constants.kvcounter_key()
+  @redis_key HostCoreTest.Constants.redis_key()
+  @redis_link HostCoreTest.Constants.default_link()
+  @redis_contract HostCoreTest.Constants.keyvalue_contract()
 
   test "can get claims" do
-    {:ok, bytes} = File.read("test/fixtures/actors/echo_s.wasm")
+    {:ok, bytes} = File.read(HostCoreTest.Constants.echo_path())
```
autodidaddict, line 11: It seems inconsistent to use the module constants on lines 4-8 but to call this function directly. For consistency, shouldn't we be using something like `@echo_path` here?

host_core/test/e2e/echo_test.exs
```diff
@@ -10,12 +10,12 @@ defmodule HostCore.E2E.EchoTest do
...
-  @echo_key "MADQAFWOOOCZFDKYEYHC7AUQKDJTP32XUC5TDSMN4JLTDTU2WXBVPG4G"
-  @httpserver_link "default"
-  @httpserver_path "test/fixtures/providers/httpserver.par.gz"
+  @echo_key HostCoreTest.Constants.echo_key()
+  @httpserver_link HostCoreTest.Constants.default_link()
+  @httpserver_path HostCoreTest.Constants.httpserver_path()
 
   test "echo roundtrip", %{:evt_watcher => evt_watcher} do
-    {:ok, bytes} = File.read("test/fixtures/actors/echo_s.wasm")
+    {:ok, bytes} = File.read(HostCoreTest.Constants.echo_path())
```
autodidaddict, line 18: Same as previous comment, should this be `@echo_path` ?

host_core/test/fixtures/actors/kvcounter_unpriv/README.md
```diff
@@ -0,0 +1,23 @@
+# KVCounter Unpriviledged
```
autodidaddict, line 1: It feels weird to me that we have an entirely different actor folder for the unprivileged actor. Is there something preventing us from just doing `make unpriv` on the original kvcounter actor makefile?

host_core/test/host_core/wasmcloud/native_test.exs
```diff
@@ -105,7 +105,7 @@ defmodule HostCore.WasmCloud.NativeTest do
 
     res =
       HostCore.WasmCloud.Native.validate_antiforgery(inv |> IO.iodata_to_binary(), [
-        "CSUPERBADKEYIAMAMALICIOUSACTOR"
+        "CMYNAMEISKEVINIAMAMALICIOUSACTOR"
```
autodidaddict, line 108: I see what you did here. :thinking:

host_core/test/test_helper.exs
```diff
@@ -1,3 +1,4 @@
 Code.require_file("support/event_watcher.exs", __DIR__)
+Code.require_file("support/constants.exs", __DIR__)
```
autodidaddict, line 2: There's a way to get around this and not need to use the `test_helper.exs` file in order to gain access to these files. No need to fix this now, I'll probably clean that up in a subsequent PR.

autodidaddict (changes requested):

host_core/test/fixtures/actors/kvcounter_unpriv/README.md
brooksmtownsend: We could do `make unpriv` if we built that in, but I feel like that gets into us making unnecessary modifications to our examples when it's only used for tests. The motivation for this is really just so that we can run `make` here and get our unpriv actor automatically built and put in place for testing.

The alternative to this would just be to go to the `examples` repo whenever we need to update this actor, comment out the `keyvalue` capability, build it, move it to fixtures, and then replace the public key in the `Constants` module. I'm fine with either way

host_core/test/host_core/wasmcloud/native_test.exs
brooksmtownsend: Oh wow who put that in there :o

host_core/test/fixtures/actors/kvcounter_unpriv/README.md
autodidaddict: For now, I think what you've got here works best since it's easily automatable with all the other samples.

host_core/test/e2e/controlinterface_test.exs
brooksmtownsend: Yep, I'll refactor this

host_core/test/e2e/echo_test.exs
brooksmtownsend: Yep, I'll refactor this

host_core/test/host_core/actors_test.exs
```diff
@@ -211,11 +202,11 @@ defmodule HostCore.ActorsTest do
         @httpserver_key,
         @httpserver_contract,
         @httpserver_link,
-        "HandleRequest",
+        "HttpServer.HandleRequest",
```
brooksmtownsend, line 205: cc @stevelr here, this was required in order to get the actor to handle the request. Is this required, or should the actor also handle `HandleRequest` on it's own?

Previously we'd invoke the actor with `wash ctl call <pubkey> HandleRequest <payload>`, I think the `HttpServer.` prefix makes this a little clunkier, but I want to make sure this is the expected behavior here.

host_core/lib/host_core/providers/provider_supervisor.ex
```diff
@@ -133,6 +133,7 @@ defmodule HostCore.Providers.ProviderSupervisor do
     |> Enum.map(fn {_d, pid, _type, _modules} ->
       provider_for_pid(pid)
     end)
+    |> Enum.filter(fn provider -> provider != nil end)
```
autodidaddict, line 136: As an FYI for the future, there's a function in the kernel called `is_nil` and you can use `reject` to reject the nils, making this a bit easier to read: 
```
Enum.reject(list, &is_nil/1)
```

autodidaddict (approved):

host_core/lib/host_core/providers/provider_supervisor.ex
brooksmtownsend: I actually like this a lot better, going to go ahead and implement it here

autodidaddict (approved):

PR #216, opened 2021-09-15
Enable verification of anti-forgery token signatures on invocations

autodidaddict:
Initial commit. There are 2 test failures but we think those may be fixture-related.


autodidaddict:
Looks like there are 2 issues happening here:

1. The `httpserver` capability provider, when building the invocation, is not building it in a way where the invocation hash on the invocation matches the generated invocation hash for comparison.. The error message is `10:36:49.895 [error] Invocation failed anti-forgery validation check: Validation of invocation/AF token failed: Invocation hash does not match signed claims hash`
2. It looks like the HTTP server always expects that the result of an invocation will always have a usable body on it, but when the `InvocationResponse` contains an error (as it does when the aft hashes don't match), the http server is failing to deserialize the empty body into an HTTP response.

@stevelr we'll need to sync up so we can figure out why the http server isn't producing a hash that the host will match.

autodidaddict:
I think I may have found why the hashes aren't matching up.. The invocation target is for an actor with a public key of `MCMF...E`, while when you look at the claims JWT on the invocation, the target public key (you can see it in the target_url field) is different:

```
{
  "jti": "Q5NkFFi9tvealXmPLkX1Or",
  "iat": 1631717746,
  "iss": "CAPX6WAAE6NL2DUTWBIGHIZG4WTZPAVJP7SPKAMCVZJXOOI56HUVWMTC",
  "sub": "35058993afa048b482f935954674c0c6",
  "wascap": {
    "target_url": "wasmbus://MBW3UGAIONCX3RIDDUGDCQIRGBQQOWS643CVICQ5EZ7SWNQPZLZTSQKU/HttpServer.HandleRequest",
    "origin_url": "wasmbus://wasmcloud/httpserver/default/VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M",
    "hash": "F4E145627F921406651801BB3840B777434B972146B9B9BC864BA0EFC7462C7F"
  }
}
```

autodidaddict:
this is almost done. Everything works except a test is failing because HTTP server capability provider isn't including the host ID in its invocations. Once [this issue](https://github.com/wasmCloud/weld/issues/34) has been resolved and merged into main, I can try this test suite again.

host_core/test/host_core/wasmcloud/native_test.exs
```diff
@@ -16,21 +16,9 @@ defmodule HostCore.WasmCloud.NativeTest do
...
-
-        {:unix, _linux} ->
-          13_572_128
-
-        {:win32, :nt} ->
-          22_161_225
-      end
-
-    assert byte_size(par.target_bytes |> IO.iodata_to_binary()) == target_bytes
+    assert byte_size(par.target_bytes |> IO.iodata_to_binary()) > 8_000_000
```
brooksmtownsend, line 21: We may want to change this later but I'm OK with it for now. Feels weird to essentially assert a minimum provider executable size.

brooksmtownsend (approved):

host_core/test/host_core/wasmcloud/native_test.exs
autodidaddict: Yeah I'm not really asserting a minimum provider executable size.. It just felt SUPER brittle asserting the file size of the provider every time. We know that the `oci-distribution` crate that we run on top of works properly, so I wanted to strip out some of the redundant and brittle-inducing assertions.

If we get a file size greater than 8MB, then we can be _reasonably sure_ that we got the right file from the OCI registry.

brooksmtownsend: Yeah agreed, it's nice to not have tests that _have_ to fail when we update. Perhaps when we get a more robust `wash par inspect` we can make actual assertions on size 😉

PR #217, opened 2021-09-15
update tests to httpserver:0.14.3

stevelr:
Signed-off-by: stevelr <steve@cosmonic.com>


autodidaddict (dismissed):

brooksmtownsend (changes requested):
As a part of this, if you could replace the `test/fixtures/providers/httpserver.par.gz` with the provider in azurecr at version `0.14.3`, that would be great.

You can simply pull it down with `wash reg pull` and drop it in the folder

brooksmtownsend (approved):

Issue #168, opened 2021-08-11
OCI reference invalid if Actor Key to OCI ref is a one-to-many relationship


brooksmtownsend:
The current code to look up an oci reference is as follows:
```
  def get_image_ref(pk) do
    case find_oci_for_pk(pk) do
      [[oci]] -> oci
      _ -> nil
    end
  end
```

This works well for when you start an actor from a remote registry. That actor's public key is correlated with the ocireference. However, in the case where more than OCI reference corresponds to the same public key, the `[[oci]]` statement will not match. See an example below for the echo actor:

```
(code)   :ets.match(:refmap_table, {:"$1", "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5"})
(result) [["wasmcloud.azurecr.io/echo:0.2.0"], ["wasmcloud.azurecr.io/echo:0.2.1"]]
```

This will instead return `nil`. At the moment elixir `ActorModule` modules _do_ have access to their own oci reference, so we should instead interrogate the `ActorModule` for its OCI reference rather than looking it up in the ets cache.

autodidaddict:
I think the behavior we want here is to only ever store one OCI reference for a given public key. When we put an OCI reference into the cache, it should over-write the previously existing reference if there was one. When we live update an actor, we replace all instances with the new version, so the "old" OCI reference is no longer valid at that point.

brooksmtownsend:
@autodidaddict What about the scenario where I have a running echo actor from both oci references? This might not be a desired outcome ATM, but feasibly this could be a scenario that we support with canary deployments, etc. This was actually how I discovered the bug, by starting an actor from `echo:0.2.0` and `echo:0.2.1`

autodidaddict:
We shouldn't be able to run two different versions of the same actor in the same host. At least that's how I've been assuming since the beginning.

PR #225, opened 2021-09-23
Closing #168. Prevents an attempt to start an actor with a conflicing…

autodidaddict:
… OCI reference.

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -156,6 +156,9 @@ defmodule HostCore.Actors.ActorSupervisor do
                  {:stop, err} ->
                    {:halt, "Error: #{err}"}
 
+                 {:error, err} ->
+                   {:halt, "Error: #{err}"}
+
                  _any ->
                    {:cont, ""}
                end
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -22,13 +22,32 @@ defmodule HostCore.Actors.ActorSupervisor do
     case HostCore.WasmCloud.Native.extract_claims(bytes) do
       {:error, err} ->
         Logger.error("Failed to extract claims from WebAssembly module")
-        {:stop, err}
+        {:error, err}
 
       {:ok, claims} ->
-        DynamicSupervisor.start_child(
-          __MODULE__,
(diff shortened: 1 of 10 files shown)


host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -22,13 +22,32 @@ defmodule HostCore.Actors.ActorSupervisor do
...
+    :ok
+  end
+
+  defp check_oci_dupes(oci, pk) do
+    if HostCore.Refmaps.Manager.ocis_for_key(pk)
+       |> Enum.reject(fn toci -> oci == toci end)
+       |> length() == 0 do
+      :ok
+    else
+      :error
```
brooksmtownsend, line 50: I think this would read better returning a truthy value, then you could just check for `if oci_is_duplicate()?` and the body of this function would simply be
```
HostCore.Refmaps.Manager.ocis_for_key(pk)
       |> Enum.reject(fn toci -> oci == toci end)
       |> length() > 0
```

This might be an opinion, and if you take a look at this and it looks more confusing just resolve this

brooksmtownsend (commented):
Just one comment, otherwise code looks good but I'm going to pull this down and ensure that this error shows in the dashboard when trying to start an actor with a different OCI ref (if you want to verify that, then that's fine too)

host_core/lib/host_core/actors/actor_supervisor.ex
autodidaddict: Oh yeah I definitely like this better. I'll push a new commit for this.

brooksmtownsend (changes requested):
This currently isn't compatible with the current implementation of `scale`, but this can be easily fixed. 

If you add the following snippet into the `reduce_while` of `scale`, this will handle the `{:error, err}` case that wasn't present before
```
                 {:error, err} ->
                   {:halt, "Error: #{err}"}
```
This change should go in just below https://github.com/wasmCloud/wasmcloud-otp/blob/bug/multipleoci/host_core/lib/host_core/actors/actor_supervisor.ex#L155

autodidaddict:
> This change should go in just below https://github.com/wasmCloud/wasmcloud-otp/blob/bug/multipleoci/host_core/lib/host_core/actors/actor_supervisor.ex#L155

Why not just make the change yourself since you know where it's supposed to go? :trollface:

brooksmtownsend (approved):
Didn't want to muddy up this great PR with my dirty commit, LGTM

Issue #223, opened 2021-09-21
Dedupe release pipeline code


thomastaylor312:
Right now we have 2 (soon to be 3 with the inclusion of Windows) fairly similar build pipelines for pushing the tarballs for each OS. These could likely be deduped with the `matrix` strategy. Based on what I see in the pipelines, we probably could divide it out into the following tasks:

- Build Docker image
- Publish to distillery
- Build and push tarballs (using `matrix`)

thomastaylor312:
Also, it would probably be good to switch out the individual compile steps for the `build-prod` target in the makefile

brooksmtownsend:
addressing in #248

PR #227, opened 2021-09-24
Adds windows tarballs to pipeline

thomastaylor312:
Please note that this does not try to consolidate any of the pipelines and the windows pipeline differs from the others. For purposes of expediency, we are doing it this way and have #223 open to track the cleanup


thomastaylor312:
Please note I tested this manually here: https://github.com/thomastaylor312/wasmcloud-otp/runs/3702450475. The linux failure is because I don't have the credentials for the docker image push

autodidaddict (approved):
Assuming this works, LGTM. I'm not sure I'm qualified enough to give this a deep critique

PR #228, opened 2021-09-24
Add the cluster issuers list to the host data structure

autodidaddict:

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -184,6 +184,7 @@ defmodule HostCore.Host do
       link_definitions: lds,
       # TODO
       env_values: %{},
+      cluster_issuers: cluster_issuers(),
       invocation_seed: cluster_seed()
     }
     |> Jason.encode!()


autodidaddict:
Issuers list field will be a list of strings (public keys). list should never be empty because the host always tries to at least use the public key of its own signing key in the issuer list.

brooksmtownsend (changes requested):
Will this cause issues with providers trying to deserialize the data structure and not knowing what to do with the `cluster_issuers` field? If so, we may want to hold off on this until we release `0.50.0` to avoid publishing all new versions of capability providers with minor version bumps

autodidaddict:
I've watched the logs and none of the providers have an issue with the extraneous field.

autodidaddict:
The tests all pass so we know they're able to get data from the host data struct

brooksmtownsend (approved):

brooksmtownsend:
Touché, I forgot that integration tests, ya know, test integration.

stevelr:
Serde deserialization ignores fields in the data stream that are not declared in the struct being deserialized into

stevelr:
Changing the providers to parse and use the cluster keys will be in 0.51

PR #229, opened 2021-09-24
fix(ci): Fixes linux tarball

thomastaylor312:
This is only a partial fix because I don't know enough about the Bocker build process yet
to untangle things properly. It is likely we'll need a new image that has rust
and elixir for building purposes, or even better, just build it on the runner
and then `ADD` the file to the elixir docker base image. However, this should
unlock having linux tarballs again


thomastaylor312:
You can see a successful run (of everything but docker) here: https://github.com/thomastaylor312/wasmcloud-otp/runs/3705343118

brooksmtownsend (approved):

PR #234, opened 2021-09-27
fix(ci): Makes sure webpack actually runs

thomastaylor312:
The assets were not being compiled into the final tarball
because we weren't webpacking them


brooksmtownsend (approved):
LGTM, I was going to suggest doing the same for Docker and for MacOS but we can do that as part of consolidating this into a matrix later on

autodidaddict (approved):

PR #248, opened 2021-10-14
Pin workflow OS versions, utilize matrix builds (WIP)

### LINKED ISSUE #223: Dedupe release pipeline code

### LINKED ISSUE #241: The host runtime fails to start
My environment is:

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic

1) Downloaded and extracted https://github.com/wasmCloud/wasmcloud-otp/releases/download/v0.50.0/x86_64-linux.tar.gz 

2) Installed NATS and other requisites as per "https://wasmcloud.dev/overview/installation/".

When I run:
`bin/wasmcloud_host foreground`

- Getting following messages:

releases/0.50.0/wasmcloud_host.sh: error while loading shared libraries: libtinfo.so.6: cannot open shared object file: No such file or directory
Unusable Erlang runtime system! This is likely due to being compiled for another system than the host is running

Subsequently:

Tried building `host_core` after installing erlang  and elixir - then I face following issue:

make build
mix deps.get
** (Mix.Config.LoadError) could not load config config/dev.exs
    ** (ArgumentError) expected config file to return keyword list, got: nil
    (mix) lib/mix/config.ex:241: Mix.Config.validate!/1
    (mix) lib/mix/config.ex:168: Mix.Config.read!/1
    (mix) lib/mix/config.ex:195: anonymous fn/2 in Mix.Config.read_wildcard!/1
    (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
    (stdlib) erl_eval.erl:878: :erl_eval.expr_list/6
    (stdlib) erl_eval.erl:404: :erl_eval.expr/5
Makefile:28: recipe for target 'deps' failed
make: *** [deps] Error 1

brooksmtownsend:
Fixes #241 
Fixes #223 

This PR does the following:
1. Consolidates the tarball build/release actions into a matrix to simplify the build process
2. Removes the `build_tarballs_manual` action in favor of allowing `wasmcloud_host` to be manually dispatched. The manual dispatch does _not_ create a release, it's only used to build the tarballs and attach them to the action
3. Adds a few extra environment variables in the docker image to suppress unnecessary warnings and automatically connect providers RPC to a `nats` container (as is the default for RPC and CTL connections)
4. Adds `linux/arm64` as a docker image platform, enabling arm devices to run the `wasmcloud_host` docker image easily

In regards to #241, this can be fixed with changing the github linux runner to `ubuntu-18.04`. _However_ this notably requires that users of ubuntu 20.04/21.04 and other equivalent linuxes must install `libtinfo5` to satisfy the shared object that's already present on 18.04.

We have a few options, and I'm looking for feedback on them
1. Build on `ubuntu-18.04`, note the `libtinfo5` dependency in our install docs
2. Build on `ubuntu-20.04`, suggest that Linux distributions that cannot install `libtinfo6` use the docker image or build from source (or upgrade to 20.04, it's not that new.)
3. Build on both, distribute 2 tarballs for linux. Because we aim to support many distributions, not just ubuntu, naming this will be challenging as debian, fedora, arch all can use this tarball.
4. Build on `ubuntu-18.04`, include the `libtinfo5` shared object as a part of our tarball, provide the symbolic link or equivalent method to enable the BEAM to use the included shared object

You can see this workflow run here: https://github.com/brooksmtownsend/wasmcloud-otp/actions/runs/1349393972
Signed-off-by: Brooks Townsend <brooksmtownsend@gmail.com>

### DIFF
--- a/.github/workflows/wasmcloud_host.yml
+++ b/.github/workflows/wasmcloud_host.yml
@@ -149,7 +149,18 @@ jobs:
     needs: build
     strategy:
       matrix:
-        os: ["ubuntu-20.04", "windows-2019", "macos-10.15"]
+        os: [ubuntu-20.04, windows-2019, macos-10.15]
+        include:
+          - os: ubuntu-20.04
+            rust-target: x86_64-unknown-linux-gnu
+            tarball-filename: x86_64-linux
+          - os: windows-2019
+            rust-target: x86_64-pc-windows-msvc
+            tarball-filename: x86_64-windows
+          - os: macos-10.15
+            rust-target: x86_64-apple-darwin
+            tarball-filename: x86_64-macos
+
     runs-on: ${{ matrix.os }}
     env:
       MIX_ENV: prod
@@ -162,22 +173,9 @@ jobs:
 
       # Setup Rust for NIF dependencies
       - uses: actions-rs/toolchain@v1
-        if: ${{ startswith(matrix.os, 'windows') }}
         with:
           toolchain: stable
-          target: x86_64-pc-windows-msvc
-          override: true
-      - uses: actions-rs/toolchain@v1
-        if: ${{ startswith(matrix.os, 'macos') }}
-        with:
-          toolchain: stable
-          target: x86_64-apple-darwin
-          override: true
-      - uses: actions-rs/toolchain@v1
-        if: ${{ startswith(matrix.os, 'ubuntu') }}
-        with:
-          toolchain: stable
-          target: x86_64-unknown-linux-gnu
+          target: ${{ matrix.rust-target }}
           override: true
 
       # Setup node for phoenix deps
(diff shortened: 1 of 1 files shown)


stevelr:
Often when there are cross version library issues like this there are "compat" packages to help bridge the gap. I searched for "libtinfo compat" and found nurses-compat-libs. There's a version 5 and a version 6. If we could add both to the tarball then a bash script could install the applicable one for the version of Ubuntu we are installing on.

Or alternately, the install script could try to install both libtinfo and ncurses-compat-libs versions based on the current version

My pref would be to link with libtinfo6 (favoring later Ubuntu) and install compat on Ubuntu 18.04 if that works.

PR #259, opened 2021-11-01
Adding support for host configuration files

autodidaddict:
Initial implementation of config file support.

A progressive merge of keys found in data files are then applied to bindings and then optionally overridden by environment variables. In order of _lowest_ to _highest_ precedence:

1. `~/.wash/host_config.json`
2. `./host_config.json` (relative to the starting host process)
3. various `.env` files - env var injection
4. real environment variables.

The JSON file keys are essentially the environment variable keys converted to lower case with the `WASMCLOUD_` prefix removed, though there are a couple of exceptions.

The files are auto-generated when a host is started without either of the files available, negating the need for developers to figure out what the keys are (the entire suite of values is dumped into the files).

This has the net effect of making it so that if I start a wasmcloud-otp host on a freshly installed machine, I will then be able to immediately start a second process that needs no further configuration in order to actively participate in the newly formed lattice.

An additional change is that the host seed and host key are no longer overridable via environment variables - hosts will now _always_ generate their own guaranteed unique key (they should've been doing this all along).


host_core/lib/host_core/config_plan.ex
```diff
@@ -5,50 +5,80 @@ defmodule HostCore.ConfigPlan do
...
 
   @prefix_var "WASMCLOUD_LATTICE_PREFIX"
-  @hostkey_var "WASMCLOUD_HOST_KEY"
-  @hostseed_var "WASMCLOUD_HOST_SEED"
   @default_prefix "default"
 
   @impl Vapor.Plan
   def config_plan do
-    {host_key, host_seed} = HostCore.WasmCloud.Native.generate_key(:server)
+    wash_config = Path.join(System.user_home(), "/.wash/host_config.json")
```
brooksmtownsend, line 15: Have we verified this canonicalizes correctly for Windows machines?

host_core/lib/host_core/files_config_provider.ex
```diff
@@ -0,0 +1,115 @@
...
+    defp decode(str, format) do
+      case format do
+        :json ->
+          Jason.decode(str)
+
+        :toml ->
+          Toml.decode(str)
+
+        :yaml ->
+          YamlElixir.read_from_string(str)
```
brooksmtownsend, line 95: What would we do without libraries.

brooksmtownsend (commented):
This largely looks good, just leaving a comment about Windows. I'm going to test it and drop an approval if it creates fine in powershell.

host_core/lib/host_core/config_plan.ex
autodidaddict: The docs for all of these functions claim to be platform independent.

brooksmtownsend (approved):
yolo

brooksmtownsend:
@autodidaddict 

```

08:43:54.326 [debug] Reading configuration file C:\Users\brooks/.wash/host_config.json

08:43:54.346 [debug] Reading configuration file host_config.json
```

It's a little confused, but it works just fine to write/read the json file 😄

autodidaddict:
> 
> It's a little confused, but it works just fine to write/read the json file smile

Yeah this is because Elixir defaults to retaining forward slashes and only uses backslashes when they come from the OS, hence the mix of back and forward (uphill and downhill?)

PR #260, opened 2021-11-04
feat(host): Adds stop host command

thomastaylor312:
This is super basic and doesn't support completely graceful shutdown, but
it should be a good start for automating host shutdowns

### DIFF
--- a/host_core/lib/host_core/control_interface/server.ex
+++ b/host_core/lib/host_core/control_interface/server.ex
@@ -202,6 +202,21 @@ defmodule HostCore.ControlInterface.Server do
     {:reply, response}
   end
 
+  # Stop Host
+  defp handle_request({"cmd", _host_id, "stop"}, body, _reply_to) do
+    # TODO: Right now this will contain a parameter for timeout. Obviously how this works currently
+    # only results in the graceful shutdowns built into the system. There may be some inflight work
+    # we want to wait for up to the timeout. We could use this library possibly so we can put in
+    # hooks: https://github.com/botsquad/graceful_stop.
+    _stop_host_command = Jason.decode!(body)
+
+    Logger.info("Received stop request for host")
+
+    HostCore.Host.purge()
+
+    :init.stop()
+  end
+
   ### AUCTIONS
   # All auctions are sent to every host within the lattice
   # so no queue subscription is used.
--- a/host_core/lib/host_core/control_interface/server.ex
+++ b/host_core/lib/host_core/control_interface/server.ex
@@ -4,6 +4,7 @@ defmodule HostCore.ControlInterface.Server do
   use Gnat.Server
 
   alias HostCore.ControlInterface.ACL
+  alias HostCore.CloudEvent
 
   def request(%{topic: topic, body: body, reply_to: reply_to}) do
     topic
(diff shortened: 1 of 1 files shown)


thomastaylor312:
Here is a video of it working: https://user-images.githubusercontent.com/4294228/140425221-d87ad723-55c5-42c4-8d55-4be8d9db898b.mov


Please note that this does not update the control interface or `wash`. That will be part of subsequent PRs

autodidaddict:
I'm a bit confused. @thomastaylor312  you have a video of it working, but the only thing in this pull request is a stub function that quite literally does nothing. Is there a commit that hasn't been pushed?

thomastaylor312:
That is because it is simple! https://github.com/wasmCloud/wasmcloud-otp/pull/260/files#diff-4d7d7351f492d39f28bde13c1b90eeaaa93b8b6d55f0fb07ab2a6614925c916eR215

Also so it is captured here in public, @autodidaddict and I talked in meatspace about this needing a host purge first. I'll add that

thomastaylor312:
Host purge added. Should be ready for review

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -202,6 +202,21 @@ defmodule HostCore.ControlInterface.Server do
     {:reply, response}
   end
 
+  # Stop Host
+  defp handle_request({"cmd", _host_id, "stop"}, body, _reply_to) do
+    # TODO: Right now this will contain a parameter for timeout. Obviously how this works currently
+    # only results in the graceful shutdowns built into the system. There may be some inflight work
+    # we want to wait for up to the timeout. We could use this library possibly so we can put in
+    # hooks: https://github.com/botsquad/graceful_stop.
+    _stop_host_command = Jason.decode!(body)
```
autodidaddict, line 211: This will crash the control interface server if the JSON decoding fails, thereby orphaning this host and making it unable to handle subsequent requests (like shutdowns). We should probably do something like `case Jason.decode(body) do` here, and `Logger.error` if the decoding fails, and then continue with the shutdown safely thereafter.

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -202,6 +202,21 @@ defmodule HostCore.ControlInterface.Server do
...
+
+    Logger.info("Received stop request for host")
```
autodidaddict, line 213: I know this is probably not part of the original scope, but I think we should create a `com.wasmcloud.lattice.host_stopped` cloud event and emit this, wait a few milliseconds, then call purge. This gives the lattice notice that a host has been _properly_ terminated (e.g. didn't crash).

You can see an example of publishing the "actor stopped" event here: https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/actors/actor_module.ex#L371

autodidaddict (changes requested):

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -202,6 +203,27 @@ defmodule HostCore.ControlInterface.Server do
...
+  # Stop Host
+  defp handle_request({"cmd", host_id, "stop"}, body, _reply_to) do
+    case Jason.decode(body) do
+      # TODO: Right now this will contain a parameter for timeout. Obviously how this works currently
+      # only results in the graceful shutdowns built into the system. There may be some inflight work
+      # we want to wait for up to the timeout. We could use this library possibly so we can put in
+      # hooks: https://github.com/botsquad/graceful_stop.
+      {:ok, _stop_host_command} ->
+        Logger.info("Received stop request for host")
+        publish_host_stopped(host_id)
```
autodidaddict, line 215: I know this is a nit, but can we publish the host stopped _after_ the purge?

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -269,4 +291,18 @@ defmodule HostCore.ControlInterface.Server do
       error: error
     })
   end
+
+  defp publish_host_stopped(host_id) do
+    prefix = HostCore.Host.lattice_prefix()
+
+    msg =
+      %{
```
autodidaddict, line 299: You can actually supply an empty map here, because the host is always provided as the `source` field on the CloudEvent, per this code https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/cloud_event.ex#L4

autodidaddict (changes requested):

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -202,6 +203,28 @@ defmodule HostCore.ControlInterface.Server do
...
+        HostCore.Host.purge()
+        publish_host_stopped()
+        # Give a little bit of time for the event to get sent before shutting down
+        :timer.sleep(100)
+
+        :init.stop()
+
+      {:error, e} ->
+        Logger.error("Unable to parse incoming stop request: #{e}")
+        failure_ack("Unable to perform live update: #{e}")
```
autodidaddict, line 224: oh noez. copy paste fail. This should be "unable to parse stop request"

autodidaddict (changes requested):

autodidaddict (approved):

host_core/lib/host_core/control_interface/server.ex
thomastaylor312: Added #265 to go back and fix this

PR #264, opened 2021-11-08
Write configuration file regardless if it exists

brooksmtownsend:
My reasoning for this is two-pronged:
1. If you are setting multiple environment variables to configure a host, it's likely that you'll want to persist those for the next time that you close your shell and start up another host. In the case that this is a production machine, and you want to avoid storing sensitive information in a file on disk, you'd have to ensure this file has been created _first_ and then set those variables according to the current logic.
2. With the presence of `host_config.json` and `~/.wash/host_config.json`, those two files should be identical to avoid confusion. If these files are only written when they don't exist, changes in the `host_config.json` relative to the host will not be reflected in `~/.wash/host_config.json` unless it's manually removed.

I sought out this solution initially because of the `wash ctx` feature, and my desire to use `~/.wash/host_config.json`'s configuration values for the default context. I think that this is a better approach regardless, and figured I'd go ahead and PR it and we can conversate here if necessary.

### DIFF
--- a/host_core/lib/host_core.ex
+++ b/host_core/lib/host_core.ex
@@ -158,11 +158,9 @@ defmodule HostCore do
   end
 
   defp write_json(config, file) do
-    if !File.exists?(file) do
-      case File.write(file, Jason.encode!(remove_extras(config))) do
-        {:error, reason} -> Logger.error("Failed to write configuration file #{reason}")
-        :ok -> Logger.info("Wrote #{inspect(file)}")
-      end
+    case File.write(file, Jason.encode!(remove_extras(config))) do
+      {:error, reason} -> Logger.error("Failed to write configuration file #{reason}")
+      :ok -> Logger.info("Wrote #{inspect(file)}")
     end
   end
 


autodidaddict:
The side-effect that I wanted to avoid by not writing the configuration file _all the time_ is that the effective configuration of a host is determined by a combination of its environment variables and the data pulled from the configuration file. If we write the configuration file every time, then what we end up with is a potential divergence from the data stored in the wash context and the data the host is actually reading from the file.

This is a peculiar one. I remember going back and forth on whether or not to "always overwrite" but I can't remember the other arguments I had against it right now.

autodidaddict:
Scenario:
1. start host with no configuration
2. both local and `~/.wash` `host_config.json` files are written
3. use `wash ctx` to create a context and set a default, overwrites `~/.wash/host_config.json`
4. run `wasmcloud_host` with multiple env vars overridding values in `host_config.json`
5. host overwrites `host_config.json`
6. host_config.json in the host's local directory _and_ in `~/.wash` now no longer match the data in the stored wash context.

brooksmtownsend:
I think I can do a better job at how `wash ctx` is implemented, and that might clear up some concerns.

Essentially, with `wash ctx` you can manage multiple contexts. I envisioned one of these contexts being managed automatically, called `host_config`, which is automatically populated by whatever is in `~/.wash/host_config.json`. If this PR is merged, this will have the net effect of always using the values in the `host_config.json` by default with `wash` (each time you run a `ctl` or `call` command the `host_config` context is automatically updated.

Besides that, you can also add additional contexts e.g. `wasmcloud_cluster_aws` for a remote AWS host, `petshop_leafnode_demo` for issuing commands to a demo cluster, but those are completely managed by the user and don't influence the values the host is launched with. My idea behind that is that you may want to issue commands to a remote cluster without having another host join it next time you launch it.

This does leave out the scenario where you may want to easily teardown your local host, switch contexts, and launch another local host with a completely new set of configuration values. I saw that as something that was out of the initial scope of the `wash ctx`.

autodidaddict:
I'm okay with this for now. We'll know more once people start using the new host whether the combination of the host_config.json file and the `wash ctx` commands has helped. We can accept PRs and issues on that rather than trying to get this perfect today.

autodidaddict (approved):

PR #266, opened 2021-11-08
feat(tests): Use the same testing in the pipeline and makefile

thomastaylor312:
Before this change, we were setting up a redis server only as part of the pipeline
while simultaneously using a docker compose file to get NATS. It is much
easier if the developer and the pipeline use the same targets. So this moves
redis to the docker compose file used for the test and removes it from the pipeline

### DIFF
--- a/host_core/test/docker-compose.yml
+++ b/host_core/test/docker-compose.yml
@@ -1,9 +1,13 @@
 version: "3"
 services:
   nats:
-    image: nats:2.3.4
+    image: nats:2
     command: "-js"
     ports:
       - "4222:4222"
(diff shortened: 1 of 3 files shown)


host_core/test/docker-compose.yml
```diff
@@ -1,9 +1,13 @@
 version: "3"
 services:
   nats:
-    image: nats:2.3.4
+    image: nats:2
```
autodidaddict, line 4: This threw me but apparently this is maintained as the latest 2.x version so that works for us.

autodidaddict (approved):

PR #268, opened 2021-11-09
Host now ACKs the host stop command

autodidaddict:
Apologies I didn't catch this before @thomastaylor312 - context switching from Rust made me forget the pattern we've been using to acknowledge commands.

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -1,6 +1,7 @@
 defmodule HostCore.Host do
   @moduledoc false
   use GenServer, restart: :transient
+  alias HostCore.CloudEvent
   require Logger
 
   # To set this value in a release, edit the `env.sh` file that is generated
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/host.ex
```diff
@@ -87,6 +88,32 @@ defmodule HostCore.Host do
     {:reply, labels, state}
   end
 
+  @impl true
+  def handle_info({:do_stop, _timeout_ms}, state) do
```
thomastaylor312, line 92: Super nit: I’ve been defining the timeout in seconds

thomastaylor312 (approved):

host_core/lib/host_core/host.ex
autodidaddict: Everything everywhere else in wasmCloud uses milliseconds. Wherever not explicitly mentioned, the assumption is ms.

PR #274, opened 2021-11-11
Adds a JetStream domain configuration parameter.

### LINKED ISSUE #273: Allow a configurable JetStream domain name in host configuration
In "standard" scenarios when we're dealing with a single NATS cluster where all nodes in that cluster are reachable, then the way we do JetStream creation of streams and the creation of ephemeral consumers works just fine.

However, when we are attempting to run in scenarios where we have multiple leaf nodes communicating across NGS (e.g. multi-cloud, multi-infra, nearly all of our multi-human demos), we run into problems. The most significant of which is that by default, `$JS.API.>` traffic is banned across the leaf node border (for lots of reasons). In order to bridge across leaf node boundaries, the hosts need to know the **domain name** of the given stream. Domains are logical groupings for JetStream resources, and _each server_ within a cluster can identify itself as part of a domain.

## Demo Scenario (Fails Today)
* Create leaf node A with JS enabled, pointing at NGS
* Create leaf node B with JS enabled, pointing at NGS
* Create stream `LATTICECACHE_default` (by starting a wasmCloud host on node A)
* Go over to node B and ask the NATS server for a stream list - reports empty
* Start a wasmCloud host on node B, it creates its own private `LATTICECACHE_default` that is empty.

## Demo Scenario (Works Tomorrow)
* Create leaf node A, JS enabled, _domain **hub**_, pointing at NGS
* Create leaf node B, JS enabled, _domain **spoke**_, pointing at NGS
* Create leaf node C, JS enabled, _domain **spoke**_, pointing at NGS
* Start a wasmCloud host with `JS_DOMAIN=hub` on node A|B|C, `LATTICECACHE_default` is created in the hub domain (in this case, on node A)
* Start a wasmCloud hsot with `JS_DOMAIN=hub` on node A|B|C, pre-existing hub cache is used for new ephemeral consumer

## Additional Notes
You can actually perform this demo without the wasmCloud hosts specifying a domain name. In each domain, you can create a _stream mirror_ from the `hub` domain. This means there's always a local stream with the identical name regardless of which NATS server domain you're in. While this is easier on the host config, it's harder on operations. As such, the wasmCloud host should support operating in both environments (implicit or explicit domains).

This issue doesn't address the edge case when all replicas of a given stream in the **hub** domain (or whatever you call it) go down. In this case, the stream will simply vanish from the lattice unless other servers/domains had been set up to mirror. This is a NATS design/architecture concern and not one that this issue thinks we need to address in the wasmCloud OTP code.

## Lamentations and Despair
Looking back on the last couple weeks worth of demos we've done, I can now see the large number of "what just happened?" failures when the distributed cache didn't work properly, including when we were prepping for KubeCon.

autodidaddict:
This closes #273. This PR includes the following:

* Re-enable the ability for a host's identity to be supplied via environment variable (`WASMCLOUD_HOST_SEED`)
* Add a new configuration value `js_domain` in the JSON config and `WASMCLOUD_JS_DOMAIN` env var
* No longer writes host configuration to `~/.wash/host_config.json` per discussions
* We now display the current consumer count on the JS stream the host is using. This can be super helpful in troubleshooting domain/leaf node issues (each active wasmCloud host represents an active ephemeral consumer)

### DIFF
--- a/host_core/lib/host_core.ex
+++ b/host_core/lib/host_core.ex
@@ -167,6 +167,14 @@ defmodule HostCore do
 
   defp write_config(config) do
     write_json(config, "./host_config.json")
+
+    case System.user_home() do
+      nil ->
+        Logger.warn("Can't check for ~/.wash host config - no user home available")
+
+      h ->
+        write_json(config, Path.join([h, "/.wash/", "host_config.json"]))
+    end
   end
 
   defp write_json(config, file) do
--- a/host_core/lib/host_core.ex
+++ b/host_core/lib/host_core.ex
@@ -81,7 +81,22 @@ defmodule HostCore do
   end
 
   defp post_process_config(config) do
-    {host_key, host_seed} = HostCore.WasmCloud.Native.generate_key(:server)
+    {host_key, host_seed} =
+      if config.host_seed == nil do
+        HostCore.WasmCloud.Native.generate_key(:server)
+      else
+        case HostCore.WasmCloud.Native.pk_from_seed(config.host_seed) do
+          {:ok, pk} ->
+            {pk, config.host_seed}
+
+          {:error, _err} ->
+            Logger.error(
+              "Failed to obtain host public key from seed: #{config.host_seed}. Using new host key."
+            )
+
+            HostCore.WasmCloud.Native.generate_key(:server)
+        end
+      end
 
     config =
       config
@@ -98,6 +113,11 @@ defmodule HostCore do
 
     hid = Hashids.encode(s, Enum.random(1..4_294_967_295))
     config = Map.put(config, :cache_deliver_inbox, "_INBOX.#{hid}")
+
+    if config.js_domain != nil && String.valid?(config.js_domain) &&
+         String.length(config.js_domain) > 1 do
+      Logger.info("Using JetStream domain: #{config.js_domain}")
+    end
 
     {def_cluster_key, def_cluster_seed} = HostCore.WasmCloud.Native.generate_key(:cluster)
     # we're generating the key, so we know this is going to work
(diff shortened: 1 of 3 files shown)


host_core/lib/host_core.ex
```diff
@@ -147,14 +167,6 @@ defmodule HostCore do
...
   defp write_config(config) do
     write_json(config, "./host_config.json")
-
-    case System.user_home() do
-      nil ->
-        Logger.warn("Can't check for ~/.wash host config - no user home available")
-
-      h ->
-        write_json(config, Path.join([h, "/.wash/", "host_config.json"]))
-    end
```
brooksmtownsend, line 157: I thought where we landed was that we shouldn't be _reading_ from `~/.wash/host_config.json` but we should still _write_ that file as that's what `wash ctx` uses for connection values

host_core/lib/host_core/config_plan.ex
```diff
@@ -12,19 +12,18 @@ defmodule HostCore.ConfigPlan do
 
   @impl Vapor.Plan
   def config_plan do
-    wash_config = Path.join(System.user_home(), "/.wash/host_config.json")
-
     [
       %Dotenv{},
       %FilesConfigProvider{
-        paths: [wash_config, "host_config.json"],
+        paths: ["host_config.json"],
```
brooksmtownsend, line 18: regarding my comment above, I think this is a good change, not reading from `~/.wash/host_config.json` anymore

brooksmtownsend (changes requested):
Only a request to still write `~/.wash/host_config.json`

host_core/lib/host_core.ex
autodidaddict: :thinking: ok. this effectively means that wash will always, by default, use the same settings as the most recently started wasmCloud host on that machine.

brooksmtownsend: Correct, that was my goal thinking of `wash` and the use of contexts primarily a developer tool. You can always "turn this off" by selecting your own default context and using those values

brooksmtownsend (approved):

PR #275, opened 2021-11-15
Bumping to v0.50.2

autodidaddict:


brooksmtownsend (changes requested):
We should bump `host_core` as well to `0.50.2`

autodidaddict:
> We should bump `host_core` as well to `0.50.2`

Woops. I was looking for that `mix.exs` module variable, and not the hardcoded string, so I missed it.

brooksmtownsend (approved):

PR #276, opened 2021-11-15
[fix] Docker release use buildx

brooksmtownsend:
This PR fixes the release action issue where the Docker image would fail to build for 2 reasons:
1. The docker config location was overwritten by the `docker login` step which was not properly tested when I used my fork (since I disabled the login steps)
2. `git` and `ca-certificates` were not available in the deps-builder image, which failed since we use a git dependency for `vapor`

### DIFF
--- a/wasmcloud_host/Dockerfile
+++ b/wasmcloud_host/Dockerfile
@@ -19,6 +19,12 @@ WORKDIR /opt/app
 # This copies our app source code into the build container
 COPY ./host_core ./host_core
 COPY ./wasmcloud_host ./wasmcloud_host
+
+# Install necessary system dependencies
+RUN apt update && \
+  DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends \
+  git
+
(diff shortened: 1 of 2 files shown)


autodidaddict (approved):

wasmcloud_host/Dockerfile
```diff
@@ -19,6 +19,14 @@ WORKDIR /opt/app
 # This copies our app source code into the build container
 COPY ./host_core ./host_core
 COPY ./wasmcloud_host ./wasmcloud_host
+
+# Install necessary system dependencies
+RUN apt update && \
```
autodidaddict, line 24: This is installing `git` in the builder image, right, not the actual runtime image? we shouldn't need git in the final execution container, right?

brooksmtownsend: Yup exactly, we only install and run what's strictly necessary in the release image

autodidaddict (approved):

PR #285, opened 2021-11-19
feat(ci): Adds auto helm chart build

thomastaylor312:
This is triggered on tag builds or on a trigger.

### DIFF
--- a/.github/workflows/wasmcloud_host.yml
+++ b/.github/workflows/wasmcloud_host.yml
@@ -268,3 +268,29 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           prerelease: false
           draft: false
+  
+  release-charts:
+    if: ${{ startswith(github.ref, 'refs/tags/') || github.event_name == 'workflow_dispatch' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.7.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.1
+        with:
+          charts_dir: wasmcloud_host
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/wasmcloud_host/chart/README.md
+++ b/wasmcloud_host/chart/README.md
@@ -20,12 +20,16 @@ recommend deploying your main NATS server with another chart.
 
 ## Running the chart
 
-Right now this chart is not available within a chart repository. We will find a place for it soon
-and update installation instructions. To use this chart, you'll need to clone the repository and
-then run from the root of the repo:
+The chart is available in a hosted Helm repository. To add the repository, run the following:
 
 ```console
-$ helm install <RELEASE_NAME> wasmcloud_host/chart
+$ helm repo add wasmcloud https://wasmcloud.github.io/wasmcloud-otp/
+```
+
+Then you can install the chart:
+
+```console
+$ helm install <RELEASE_NAME> wasmcloud/wasmcloud-host
 ```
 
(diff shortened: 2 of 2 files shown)


wasmcloud_host/chart/README.md
```diff
@@ -20,12 +20,16 @@ recommend deploying your main NATS server with another chart.
...
 ## Running the chart
 
-Right now this chart is not available within a chart repository. We will find a place for it soon
-and update installation instructions. To use this chart, you'll need to clone the repository and
-then run from the root of the repo:
+The chart is available in a hosted Helm repository. To add the repository, run the following:
 
 ```console
-$ helm install <RELEASE_NAME> wasmcloud_host/chart
+$ helm repo add wasmcloud https://wasmcloud.github.io/wasmcloud-otp/
```
stevelr, line 26: that url returns 404

.github/workflows/wasmcloud_host.yml
```diff
@@ -268,3 +268,29 @@ jobs:
...
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.7.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.1
+        with:
+          charts_dir: wasmcloud_host
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
stevelr, line 296: nit: I think yaml is supposed to have newline at end of file.

.github/workflows/wasmcloud_host.yml
```diff
@@ -268,3 +268,29 @@ jobs:
...
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.7.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.1
```
stevelr, line 292: Should we add some chart testing like this?
 https://github.com/marketplace/actions/helm-chart-testing

stevelr (changes requested):
Only two issues:
- it would be great to add chart testng. if that's not easy to do, at a minimum add a yaml linter.
- check the bad url
otherwise looks good.

.github/workflows/wasmcloud_host.yml
```diff
@@ -268,3 +268,29 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           prerelease: false
           draft: false
+  
+  release-charts:
+    if: ${{ startswith(github.ref, 'refs/tags/') || github.event_name == 'workflow_dispatch' }}
```
stevelr, line 273: to think about for future: Do these charts apply only for our current set of releases which include wasmcloud_host? If we started releasing host_core separately, would we need new charts? If yes, we'll need to change this condition to run the workflow only for wasmcloud_host release tags.

wasmcloud_host/chart/README.md
autodidaddict: confirmed this is a 404 for me, too

autodidaddict (changes requested):
Looks fine to me other than the one repo URL that doesn't resolve.

wasmcloud_host/chart/README.md
thomastaylor312: Because it won't exist until I publish for the first time

.github/workflows/wasmcloud_host.yml
thomastaylor312: Yes it is. Will fix

.github/workflows/wasmcloud_host.yml
thomastaylor312: Yep, that is correct

.github/workflows/wasmcloud_host.yml
thomastaylor312: Let me add that as a follow up issue as it is a separate pipeline and concern from this PR.

thomastaylor312: https://github.com/wasmCloud/wasmcloud-otp/issues/288

autodidaddict (approved):

thomastaylor312:
@autodidaddict @stevelr Should be ready for re-review

stevelr (approved):

PR #286, opened 2021-11-20
Hotwatching Actors

brooksmtownsend:
This PR implements hot-reloading for actors running on a local wasmCloud host. "hot-reloading" here means that you can load an actor from a file, and anytime that file is modified (e.g. signing with new claims, changed logic), we stop all running instances of that actor and start new instances from the updated file.

This is done in a similar way to the current `From File` process, however, we require you to give the absolute path (or the path relative to where you are running your host) to the actor module. This is unfortunate, but required since uploading files in a browser does not give you the absolute path (security implications, browsers can't know things like this about the host machine). This is relatively frictionless currently, as the `make target-path-abs` will give you the absolute path to the file easily and is included in our project template (TYSM @stevelr)

This is implemented with a GenServer that maintains a map of paths to actor IDs, and registers interest with the FileSystem GenServer for modification events. This is roughly the same process that Phoenix uses to watch files for the live-reloading feature.


wasmcloud_host/lib/wasmcloud_host_web/live/components/start_actor_component.ex
```diff
@@ -115,23 +137,35 @@ defmodule StartActorComponent do
...
       end
 
     ~L"""
     <form class="form-horizontal" phx-submit="<%= modal_id %>" phx-change="validate" phx-target="<%= @myself %>">
       <input name="_csrf_token" type="hidden" value="<%= Phoenix.Controller.get_csrf_token() %>">
+      <%= if assigns.id == :start_actor_file_hotreload_modal do %>
+      <div class="form-group row" phx-drop-target="<%= @uploads.actor.ref %>">
+        <label class="col-md-3 col-form-label" for="file-input">Path</label>
+        <div class="col-md-9">
+          <input class="form-control" id="text-input" type="text" name="path"
```
stevelr, line 153: - I'm curious - why a text field and not `live_file_input`, as used in line 164 for upload actor, which provides drag & drop capability and is made for file paths?

stevelr (approved):
super useful feature!

- not a blocker for this PR, but would it be hard to do the same for providers also?

brooksmtownsend:
> super useful feature!
> 
> 
> 
> - not a blocker for this PR, but would it be hard to do the same for providers also?

I can do some investigation here, I don't think it would be too difficult but the feedback loop may be longer due to creating the PAR+different stop start mechanisms, let me look into this and I'll report back (FWIW it could be worth it to do in another PR regardless)

wasmcloud_host/lib/wasmcloud_host_web/live/components/start_actor_component.ex
brooksmtownsend: The only reason here is that live_file_input ends up creating a copy of the actor in the temp directory, which of course isn't repeated for actor changes, which we need to do a seamless watch for file events. If I missed a way to do that in Phoenix that ends up getting around the browser restriction that prevents this, I'd be happy to change that. Originally I envisioned this feature being a checkbox on the "From File" modal.

stevelr:
agreed - provider update shouldn't be added to this PR.

autodidaddict (approved):

PR #291, opened 2021-11-23
Acknowledge start provider and actor early, clean up error handling and edge cases in control interface server

autodidaddict:


host_core/lib/host_core/control_interface/server.ex
```diff
@@ -117,104 +129,153 @@ defmodule HostCore.ControlInterface.Server do
...
-        {:reply, failure_ack("Failed to start actor: #{e}")}
+    with {:ok, start_actor_command} <- Jason.decode(body),
+         true <-
+           Map.has_key?(start_actor_command, "actor_ref") do
+      Task.start(fn ->
+        case HostCore.Actors.ActorSupervisor.start_actor_from_oci(
+               start_actor_command["actor_ref"]
+             ) do
+          {:ok, _pid} ->
+            Logger.debug("Completed request to start actor")
```
thomastaylor312, line 140: Nit: Should this log contain the actor_ref?

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -117,104 +129,153 @@ defmodule HostCore.ControlInterface.Server do
...
     case Jason.decode(body) do
       # TODO: Right now this will contain a parameter for timeout. Obviously how this works currently
       # only results in the graceful shutdowns built into the system. There may be some inflight work
       # we want to wait for up to the timeout. We could use this library possibly so we can put in
       # hooks: https://github.com/botsquad/graceful_stop.
       {:ok, stop_host_command} ->
-        Logger.info("Received stop request for host")
-        Process.send_after(HostCore.Host, {:do_stop, stop_host_command["timeout"]}, 100)
-        {:reply, success_ack()}
+        if host_id != HostCore.Host.host_key() do
```
thomastaylor312, line 272: Should these be `==` instead of `!=`?

thomastaylor312 (changes requested):
One minor blocker that could just be me reading the code incorrectly

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -229,45 +290,59 @@ defmodule HostCore.ControlInterface.Server do
...
-        provider_ref: auction_request["provider_ref"],
-        link_name: auction_request["link_name"],
-        constraints: auction_request["constraints"],
-        host_id: HostCore.Host.host_key()
-      }
-
-      {:reply, Jason.encode!(ack)}
+    with {:ok, auction_request} <- Jason.decode(body),
+         true <-
+           ["constraints", "provider_ref"]
```
brooksmtownsend, line 323: I went back and forth on if `link_name` should be here as well, initially I thought it should but if my assumptions are correct it shouldn't be needed:
1. `auction_request["link_name"]` without one present resolves to `nil`
1. `link_name` is optional on the control interface, so being specified as `link_name: nil` here simply ends up being the equivalent of `None` in Rust

brooksmtownsend (commented):
Agreed with Taylor's points, and I have an observation that you can resolve if my assumptions are correct

host_core/lib/host_core/control_interface/server.ex
autodidaddict: I actually had intended it to be there but I missed it. Instead I think I'll make it default to `default` rather than nil.

thomastaylor312 (approved):

PR #293, opened 2021-11-24
Emits human-friendly relative times for claims expiration and not-before dates

autodidaddict:
Adds the human-friendly (e.g. the phrases used by wash during claims `inspect`) time-relative phrases for when claims expire and when they become valid. The events affected are actor_started and provider_started.


thomastaylor312 (approved):
All the references of `human` really add a nice robotic touch to our human impersonations

PR #295, opened 2021-11-29
Fix/hotreload events

brooksmtownsend:
This PR fixes 2 bugs in hot reloading:

Different operating systems have different filesystem events, so my logic wasn't entirely correct. For example, on Linux there is an event for `:modified` AND an event that combines `:modified` and `:closed`, which was how I based my "done editing actor" logic before. Now, we reload whenever an actor is `:modified`

Not all OSes emit the `:closed` event, and often one run of `make` to update an actor will result in multiple modified events. With my testing on the M1 mac, I would get 5 rapid-fire events of modified for one change. The new logic to handle this requests a reload of an actor after a specified delay and sets a value in state to indicate that the actor is pending a reload so we don't reload each instance 5 times.

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
+++ b/wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
@@ -27,7 +27,6 @@ defmodule WasmcloudHost.ActorWatcher do
   def handle_info({:file_event, _watcher_pid, {path, events}}, state) do
     # Modified is emitted on Mac, Windows, and Linux when a file is changed
     if :modified in events do
-      {:ok, _bytes} = File.read(path)
       actor_map = Map.get(state, path, %{})
       actor_id = Map.get(actor_map, :actor_id, "")
       is_reloading = Map.get(actor_map, :is_reloading, false)
--- a/wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
+++ b/wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
@@ -1,7 +1,7 @@
 defmodule WasmcloudHost.ActorWatcher do
   use GenServer
 
-  @reload_delay 1_000
+  @reload_delay_ms 1_000
 
   def start_link(args) do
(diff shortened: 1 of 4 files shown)


wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
```diff
@@ -1,6 +1,8 @@
 defmodule WasmcloudHost.ActorWatcher do
   use GenServer
 
+  @reload_delay 1_000
```
stevelr, line 4: preferably `@reload_delay_ms`

stevelr (approved):

wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
```diff
@@ -23,9 +25,12 @@ defmodule WasmcloudHost.ActorWatcher do
   end
 
   def handle_info({:file_event, _watcher_pid, {path, events}}, state) do
-    if :modified in events and :closed in events do
-      {:ok, bytes} = File.read(path)
-      actor_id = Map.get(state, path, "")
+    # Modified is emitted on Mac, Windows, and Linux when a file is changed
+    if :modified in events do
+      {:ok, _bytes} = File.read(path)
```
thomastaylor312, line 30: Minor nit: What is this read asserting? Is it to check that the file actually exists? I'd recommend using the equivalent of stat-ing the file rather than reading the whole module in and throwing away the bytes. If people end up doing larger modules, this would take a while every time a reload is triggered

thomastaylor312 (dismissed):
Approved with 1 minor nit

wasmcloud_host/lib/wasmcloud_host/actor_watcher.ex
brooksmtownsend: We actually don't need that at all, it was an artifact of refactoring. I'll remove it

autodidaddict (approved):

PR #296, opened 2021-11-30
fix(chart): Fixes missing secret

thomastaylor312:
The `or` clause for the secret was only checking if one of the possible
values was set. This now fixes it to make sure that any of them will
trigger secret creation


brooksmtownsend (approved):
It's been a while since I wrote a helm chart... but you're supposed to be good with this right? LGTM

PR #299, opened 2021-12-01
Generate secret key base for user when building docker image

brooksmtownsend:
This is a small fix for the Makefile, but when users currently do `make build-image` to build a docker image it will autofill with an empty `SECRET_KEY_BASE`, which unfortunately will cause failures in the web dashboard all the way at runtime. This will use `mix` to generate a secret if it's not supplied.

I also considered just running `mix phx.gen.secret` and committing the dummy value on my own, though it felt a little wrong. Happy to do this if we are OK with committing a dummy secret in exchange for not requiring an elixir toolchain to build an image.

### DIFF
--- a/wasmcloud_host/Makefile
+++ b/wasmcloud_host/Makefile
@@ -6,6 +6,7 @@ BUILD ?= `git rev-parse --short HEAD`
 DOCKERFILE ?= ./wasmcloud_host/Dockerfile
 SKIP_PHOENIX ?= false
 TAG ?= latest
+SECRET_KEY_BASE ?= $(shell mix phx.gen.secret)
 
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
I'm fine with this, but maybe @autodidaddict should take a look because I don't fully understand all the implications of `SECRET_KEY_BASE`

stevelr (approved):
agree that generating unique keys in the Makefile is right - we don't want secrets in git/source tree.

autodidaddict:
For context, the secret key base environment variable is used by Phoenix to encrypt and verify cookies, session tokens, etc.

autodidaddict (approved):

PR #301, opened 2021-12-02
Adds image ref to various events

autodidaddict:
Also adds `contract_id` to provider stopped event.


brooksmtownsend (approved):
LGTM, I don't think this will break anything either since our deserialize logic just ignores extra fields.

PR #305, opened 2021-12-02
Fix control interface inventory

brooksmtownsend:
This PR introduces a few fixes:
1. Modified the shape of the actor list in the `host inventory` query to match the control interface structure [here](https://docs.rs/wasmcloud-interface-lattice-control/0.2.2/wasmcloud_interface_lattice_control/struct.ActorInstance.html)
2. removed the unused `${{ env.working-directory }}` from the release action that was causing the GH release to fail
3. Added the `name` and `revision` information to the Actor and Provider inventory queries
4. Bumped to `0.51.1` as this PR fixes a bug, no breaking change

### DIFF
--- a/host_core/lib/host_core/control_interface/acl.ex
+++ b/host_core/lib/host_core/control_interface/acl.ex
@@ -4,30 +4,44 @@ defmodule HostCore.ControlInterface.ACL do
   def all_actors() do
     HostCore.Actors.ActorSupervisor.all_actors()
     |> Enum.flat_map(fn {id, pids} ->
+      name = get_name(id)
       revision = get_revision(id)
+
+      instances =
+        pids
+        |> Enum.map(fn pid ->
+          %{
+            annotations: %{},
+            instance_id: HostCore.Actors.ActorModule.instance_id(pid),
+            revision: revision
+          }
+        end)
 
       pids
       |> Enum.map(fn pid ->
         %{
           id: id,
-          revision: revision,
           image_ref: HostCore.Actors.ActorModule.ociref(pid),
-          instance_id: HostCore.Actors.ActorModule.instance_id(pid)
+          name: name,
+          instances: instances
         }
       end)
     end)
   end
 
   def all_providers() do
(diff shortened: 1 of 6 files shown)


host_core/lib/host_core/control_interface/acl.ex
```diff
@@ -4,30 +4,44 @@ defmodule HostCore.ControlInterface.ACL do
...
     HostCore.Actors.ActorSupervisor.all_actors()
     |> Enum.flat_map(fn {id, pids} ->
+      name = get_name(id)
       revision = get_revision(id)
 
+      instances =
+        pids
+        |> Enum.map(fn pid ->
+          %{
+            annotations: %{},
```
thomastaylor312, line 14: Should this be getting annotations from somewhere?

host_core/lib/host_core/control_interface/acl.ex
```diff
@@ -4,30 +4,44 @@ defmodule HostCore.ControlInterface.ACL do
...
       %{
         id: pk,
-        link_name: link,
-        revision: 0,
         image_ref: HostCore.Providers.ProviderModule.ociref(pid),
-        instance_id: instance_id
+        link_name: link,
+        name: name,
+        instance_id: instance_id,
+        revision: revision
```
thomastaylor312, line 44: Suuuuuper nit: Technically adding this is not in scope for a patch release, but given that we are pre-release it is probably just fine

thomastaylor312 (approved):

autodidaddict (approved):
Revision should actually come from the individual actor pid but I'm not sure it's of significant impact (yet)

host_core/lib/host_core/control_interface/acl.ex
brooksmtownsend: I see support for annotations on the control interface commands for starting an actor but we don't actually use them ATM.

autodidaddict: Annotations aren't fully supported yet.

PR #310, opened 2021-12-06
Terminate all instances of an actor with a count of zero

brooksmtownsend:
When a command comes in to terminate `0` instances of an actor, we interpret that as a request to terminate all instances. This PR implements that functionality


autodidaddict (approved):
Love the use of the guard clauses here.

PR #311, opened 2021-12-06
Moved arm64 build to its own workflow step

brooksmtownsend:
This PR moves the 3ish hour ARM docker image build step into its own contained (ha) step that won't impede the release of tarballs or the amd64 docker image.

I've also tagged the arm image separately which seems like a common pattern from doing some research. I'm currently investigating if we can avoid doing this separate tag and just push to the same tag with 2 platforms, but I don't want to overwrite the amd64 image doing so.

Signed-off-by: Brooks Townsend <brooks@cosmonic.com>


brooksmtownsend:
closing this as there is a better solution to this problem

PR #312, opened 2021-12-07
Ensured actors and providers accurately return errors

brooksmtownsend:
Previously when starting an actor from the control interface, if the OCI bytes failed to download we would return `{:stop, err}`. This was inconsistent with other failure modes, so I changed that to `{:error, err}`.

Additionally, some failures for starting providers from the control interface would simply return the result of logging the error, which was `:ok`. I added an additional line to properly return `{:error, err}` here.

Also bump to `0.51.4` for these changes.

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -53,7 +53,7 @@ defmodule HostCore.Actors.ActorSupervisor do
          ) do
       {:error, err} ->
         Logger.error("Failed to download OCI bytes for #{oci}")
-        {:stop, err}
+        {:error, err}
 
       {:ok, bytes} ->
         start_actor(bytes |> IO.iodata_to_binary(), oci)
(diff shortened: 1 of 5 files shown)


host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -53,7 +53,7 @@ defmodule HostCore.Actors.ActorSupervisor do
          ) do
       {:error, err} ->
         Logger.error("Failed to download OCI bytes for #{oci}")
-        {:stop, err}
+        {:error, err}
```
autodidaddict, line 56: yeah this is crazy that we didn't do anything about this earlier. Returning `{:stop, ...}` actually stops the supervisor, thereby preventing any other actor from being able to start on this host.

autodidaddict (approved):

host_core/lib/host_core/actors/actor_supervisor.ex
brooksmtownsend: Let's hope that we only returned `{:stop` to a place that didn't do that 😮

PR #314, opened 2021-12-07
Updated dashboard heartbeat shape

### LINKED ISSUE #304: [BUG] Missing claims for actors can crash washboard
Because of the structure of [this line](https://github.com/wasmCloud/wasmcloud-otp/blob/main/wasmcloud_host/lib/wasmcloud_host_web/live/page_live.html.heex#L65), if an actor's claims is not properly loaded in the washboard then it will cause an elixir error. This is because the `Enum.find` will result in `nil` if the actor isn't found, and you cannot do `nil |> elem(1)` because that expects a tuple. 

This was actually already fixed for providers [here](https://github.com/wasmCloud/wasmcloud-otp/blob/main/wasmcloud_host/lib/wasmcloud_host_web/live/page_live.html.heex#L135), by simply returning a tuple `("", %{})` which will appropriately not render without an error. The same process should be done for the actor claims logic.

brooksmtownsend:
This PR updates the dashboard to use the new heartbeat format introduced in #309. 

This also fixes #304 and will not cause a dashboard crash in the case of missing claims (instead, the actor name will show as `N/A`)

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
+++ b/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
@@ -257,17 +257,18 @@ defmodule WasmcloudHost.Lattice.StateMonitor do
     actor_map =
       actors
       |> Enum.reduce(%{}, fn actor, actor_map ->
-        actor_id = Map.get(actor, "actor")
-        existing_actor = current_host |> Map.get(:actors, %{}) |> Map.get(actor_id)
+        actor_id = Map.get(actor, "public_key")
 
-        if existing_actor != nil && Map.get(existing_actor, :count) == Map.get(actor, "instances") do
-          Map.put(actor_map, actor_id, existing_actor)
-        else
-          Map.put(actor_map, actor_id, %{
-            count: Map.get(actor, "instances"),
-            status: "Awaiting"
-          })
-        end
+        actor_info = Map.get(actor_map, actor_id, %{})
+        count = Map.get(actor_info, :count, 0) + 1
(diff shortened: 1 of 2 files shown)


wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
```diff
@@ -257,17 +257,18 @@ defmodule WasmcloudHost.Lattice.StateMonitor do
...
-        if existing_actor != nil && Map.get(existing_actor, :count) == Map.get(actor, "instances") do
-          Map.put(actor_map, actor_id, existing_actor)
-        else
-          Map.put(actor_map, actor_id, %{
-            count: Map.get(actor, "instances"),
-            status: "Awaiting"
-          })
-        end
+        actor_info = Map.get(actor_map, actor_id, %{})
+        count = Map.get(actor_info, :count, 0) + 1
```
autodidaddict, line 263: There's no `count` field on the list of actors that comes with the new heartbeat. The new heartbeat `actors` field is a list of `%{public_key: _, instance_id: _}` , as shown here https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/heartbeat_emitter.ex#L41

If you want the count of each unique actor on the host then you'll need to do a `Enum.group_by` on the actor's public key and count the instances (values).

autodidaddict (changes requested):
While this might prevent errors, I don't think this accurately obtains the actor count.

autodidaddict (approved):

PR #319, opened 2021-12-18
Adding support for friendly host names. Fixing bug with writing config file

autodidaddict:
This PR adds support for randomly generated friendly host names. These are not guaranteed to be globally unique, but it's a pretty good chance they'll be unique within a lattice (and there are no real consequences if there's a .001% chance duplicate).

Additionally, while I was in here I fixed a bug that came from assuming that the `~/.wash` directory exists prior to host launch.

```
15:24:22.149 [info]  Host NBLVEPX5AF35DKQVC6MBHGWN4WXIVWCAZX6MHOF6T6OAXH4AEWUXZPTL (misty-ocean-3359) started.

```

### DIFF
--- a/host_core/lib/host_core.ex
+++ b/host_core/lib/host_core.ex
@@ -181,9 +181,14 @@ defmodule HostCore do
   end
 
   defp write_json(config, file) do
-    case File.write(file, Jason.encode!(remove_extras(config))) do
-      {:error, reason} -> Logger.error("Failed to write configuration file #{reason}")
-      :ok -> Logger.info("Wrote #{inspect(file)}")
+    with :ok <- File.mkdir_p(Path.dirname(file)) do
+      case File.write(file, Jason.encode!(remove_extras(config))) do
+        {:error, reason} -> Logger.error("Failed to write configuration file #{file}: #{reason}")
+        :ok -> Logger.info("Wrote #{inspect(file)}")
+      end
+    else
+      {:error, posix} ->
+        Logger.error("Failed to create path to config file #{file}: #{posix}")
     end
   end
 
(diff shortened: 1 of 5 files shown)


host_core/lib/host_core.ex
```diff
@@ -181,9 +181,14 @@ defmodule HostCore do
   end
 
   defp write_json(config, file) do
-    case File.write(file, Jason.encode!(remove_extras(config))) do
-      {:error, reason} -> Logger.error("Failed to write configuration file #{reason}")
-      :ok -> Logger.info("Wrote #{inspect(file)}")
+    with :ok <- File.mkdir_p(Path.dirname(file)) do
+      case File.write(file, Jason.encode!(remove_extras(config))) do
```
brooksmtownsend, line 185: It's probably unnecessary, but have we considered adding `Jason.encode(_)` to the `with` statement to safely handle any encoding failures?

I can see this being more useful in regards to decoding unknown input, but it might help here too

brooksmtownsend (commented):
Comment/question, largely LGTM to me though 🙂 

The randomness of these names actually gives us something else fun, we should announce a contest to get a free sticker / T shirt for the first person to launch two hosts that have the same friendly name

host_core/lib/host_core.ex
autodidaddict: Most of the time I'd agree, but since we're the ones emitting the struct, and we know the output of `remove_extras` will always produce a JSON-encodable map, we're safe using the `!` here.

brooksmtownsend (approved):

autodidaddict:
> Comment/question, largely LGTM to me though slightly_smiling_face
> 
> The randomness of these names actually gives us something else fun, we should announce a contest to get a free sticker / T shirt for the first person to launch two hosts that have the same friendly name

Statistically, such a winner would likely never be found :(

brooksmtownsend:
> Statistically, such a winner would likely never be found :(

Ah we could hit matching adjectives and nouns but likely not an ID :(

PR #320, opened 2021-12-20
Emit invocation success and failure events

autodidaddict:
This PR adds support for publishing invocation success and failure events. These events are directional and so preserve the source and destination of the invocation. This will emit events for comms between actors and providers as well as actor<->actor comms.

Sample events:
```
{
    "data": {
        "bytes": 51,
        "dest": {
            "contract_id": "wasmcloud:httpserver",
            "link_name": "default",
            "public_key": "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M"
        },
        "operation": "HttpServer.HandleRequest",
        "source": {
            "contract_id": "",
            "link_name": "",
            "public_key": "system"
        }
    },
    "datacontenttype": "application/json",
    "id": "b2284728-d463-445d-ac7f-70174255b795",
    "source": "NDJ3Z7CBLV3UCPCENHQY34LG3XB7J766DBVG6MM6R34SGN2QDYEYN27K",
    "specversion": "1.0",
    "time": "2021-12-20T20:38:47.755590Z",
    "type": "com.wasmcloud.lattice.invocation_succeeded"
}

{
    "data": {
        "bytes": 8,
        "dest": {
            "contract_id": null,
            "link_name": null,
            "public_key": "MBMOM2EZZICFYM4KATRMH2JUO5QWE3YWCHGFZVRQQ2SQI4I5BKWIGMBS"
        },
        "operation": "Ping",
        "source": {
            "contract_id": null,
            "link_name": null,
            "public_key": "MDCX6E7RPUXSX5TJUD34CALXJJKV46MWJ2BUJQGWDDR3IYRJIWNUQ5PN"
        }
    },
    "datacontenttype": "application/json",
    "id": "21900ada-daec-41bc-a289-951b8dab6b89",
    "source": "NDJ3Z7CBLV3UCPCENHQY34LG3XB7J766DBVG6MM6R34SGN2QDYEYN27K",
    "specversion": "1.0",
    "time": "2021-12-20T20:38:49.941137Z",
    "type": "com.wasmcloud.lattice.invocation_succeeded"
}
```

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -367,7 +367,7 @@ defmodule HostCore.Actors.ActorModule do
           link_name: Map.get(target, "link_name")
         },
         operation: inv["operation"],
-        bytes: byte_size(inv["msg"])
+        bytes: byte_size(Map.get(inv, "msg", ""))
       }
       |> CloudEvent.new(evt_type)
 
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -172,50 +172,55 @@ defmodule HostCore.Actors.ActorModule do
     Logger.debug("Received invocation on #{topic}")
     iid = Agent.get(agent, fn content -> content.instance_id end)
 
-    ir =
+    {ir, inv} =
       with {:ok, inv} <- Msgpax.unpack(body) do
         case HostCore.WasmCloud.Native.validate_antiforgery(body, HostCore.Host.cluster_issuers()) do
           {:error, msg} ->
             Logger.error("Invocation failed anti-forgery validation check: #{msg}")
 
-            %{
-              msg: nil,
-              invocation_id: inv["id"],
-              error: msg,
-              instance_id: iid
-            }
+            {%{
+               msg: nil,
+               invocation_id: inv["id"],
+               error: msg,
+               instance_id: iid
+             }, inv}
 
           _ ->
             case perform_invocation(agent, inv["operation"], inv["msg"]) do
               {:ok, response} ->
-                %{
-                  msg: response,
-                  invocation_id: inv["id"],
-                  instance_id: iid
-                }
+                {%{
+                   msg: response,
+                   invocation_id: inv["id"],
+                   instance_id: iid
+                 }, inv}
 
               {:error, error} ->
                 Logger.error("Invocation failure: #{error}")
 
-                %{
-                  msg: nil,
-                  error: error,
-                  invocation_id: inv["id"],
-                  instance_id: iid
-                }
+                {%{
+                   msg: nil,
+                   error: error,
+                   invocation_id: inv["id"],
+                   instance_id: iid
+                 }, inv}
             end
         end
       else
         _ ->
-          %{
-            msg: nil,
-            invocation_id: "",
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -336,6 +341,40 @@ defmodule HostCore.Actors.ActorModule do
...
+          contract_id: Map.get(origin, "contract_id"),
+          link_name: Map.get(origin, "link_name")
+        },
+        dest: %{
+          public_key: target["public_key"],
+          contract_id: Map.get(target, "contract_id"),
+          link_name: Map.get(target, "link_name")
+        },
+        operation: inv["operation"],
+        bytes: byte_size(inv["msg"])
```
brooksmtownsend, line 370: This looks like it will fail if `inv["msg"]` is `nil`, we might want to do a `Map.get(inv, "msg", "")` here
```
byte_size(nil)
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not a bitstring

    :erlang.byte_size(nil)
```

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -299,15 +300,83 @@ defmodule HostCore.WebAssembly.Imports do
       )
       |> perform_rpc_invoke(target_subject)
 
-    case unpack_invocation_response(invocation_res) do
-      {1, :host_response, msg} ->
-        Agent.update(agent, fn state -> %State{state | host_response: msg} end)
-        1
+    unpacked = unpack_invocation_response(invocation_res)
```
brooksmtownsend, line 303: nit: we don't use `unpacked` anywhere but the `case` so this function call can probably be used in the `case` statement instead

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -299,15 +300,83 @@ defmodule HostCore.WebAssembly.Imports do
...
+      if res == 1 do
+        "invocation_succeeded"
+      else
+        "invocation_failed"
+      end
+
+    prefix = HostCore.Host.lattice_prefix()
+
+    msg =
+      %{
```
brooksmtownsend, line 352: This large nested map structure shows up in 2 places, I think it would be cleaner if we had a helper function that could take all of the parameters and form the structure just to avoid having to change it in both places.

brooksmtownsend (changes requested):

host_core/lib/host_core/web_assembly/imports.ex
autodidaddict: yeah I'd refactored this thing a bunch of times, so there's some leftover goo.

autodidaddict: so I agree that it would ordinarily be a good idea to collapse this into one.. but when you look at the sources of data for each of the two uses of this structure, they're very different. By the time I massage both inputs into a central way of producing this structure, it ends up being more code than the 2 separate uses.

brooksmtownsend: Yeah that's fair, I'm fine with just leaving as-is. We'll just have to keep that in mind for any future updates to the invocation event structure

brooksmtownsend (approved):

PR #323, opened 2021-12-27
feat(host_core): Adds experimental support for bindle

thomastaylor312:
This PR adds experimental support for [Bindle](https://github.com/deislabs/bindle)
to wasmCloud. To be very clear: THIS IS EXPERIMENTAL/could change at any time and its
use is not currently recommended in production. You can now pass a bindle ID
prefaced with `bindle://` and the host will attempt to use a bindle server.

If this PR is merged, I will create a follow up PR for the wasmCloud docs with
example invoices and instructions on how to use it

### DIFF
--- a/host_core/lib/host_core/control_interface/server.ex
+++ b/host_core/lib/host_core/control_interface/server.ex
@@ -130,14 +130,22 @@ defmodule HostCore.ControlInterface.Server do
 
   # Launch Actor
   # %{"actor_ref" => "wasmcloud.azurecr.io/echo:0.12.0", "host_id" => "Nxxxx"}
+  # %{"actor_ref" => "bindle://example.com/echo/0.12.0", "host_id" => "Nxxxx"}
   defp handle_request({"cmd", _host_id, "la"}, body, _reply_to) do
     with {:ok, start_actor_command} <- Jason.decode(body),
          true <-
            Map.has_key?(start_actor_command, "actor_ref") do
       Task.start(fn ->
-        case HostCore.Actors.ActorSupervisor.start_actor_from_oci(
-               start_actor_command["actor_ref"]
-             ) do
+        res =
+          if String.starts_with?(start_actor_command["actor_ref"], "bindle://") do
+            HostCore.Actors.ActorSupervisor.start_actor_from_bindle(
+              start_actor_command["actor_ref"]
+            )
+          else
+            HostCore.Actors.ActorSupervisor.start_actor_from_oci(start_actor_command["actor_ref"])
+          end
+
+        case res do
           {:ok, _pid} ->
             Logger.debug("Completed request to start actor #{start_actor_command["actor_ref"]}")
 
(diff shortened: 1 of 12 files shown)


host_core/lib/host_core/control_interface/server.ex
```diff
@@ -130,14 +130,22 @@ defmodule HostCore.ControlInterface.Server do
...
     with {:ok, start_actor_command} <- Jason.decode(body),
          true <-
            Map.has_key?(start_actor_command, "actor_ref") do
       Task.start(fn ->
-        case HostCore.Actors.ActorSupervisor.start_actor_from_oci(
-               start_actor_command["actor_ref"]
-             ) do
+        res =
+          if String.starts_with?(start_actor_command["actor_ref"], "bindle://") do
+            HostCore.Actors.ActorSupervisor.start_actor_from_bindle(
```
autodidaddict, line 141: cleanliness hint.. you can add the following to the top of the file so that you can remove all the module names from this call and just invoke `start_actor_from_bindle` (and the others) as though they were module-local:

```elixir
import HostCore.Actors.ActorSupervisor, only: [start_actor_from_bindle: 1, start_actor_from_oci:1]
import HostCore.Providers.ProviderSupervisor, only: [start_provider_from_bindle:1, start_provider_From_oci:1]
```

host_core/native/hostcore_wasmcloud_native/src/lib.rs
```diff
@@ -70,10 +75,123 @@ rustler::init!(
         par_cache_path,
         detect_core_host_labels,
         pk_from_seed,
+        get_provider_bindle,
+        get_actor_bindle,
     ],
     load = load
 );
 
+#[rustler::nif]
```
autodidaddict, line 84: Retrieving bindles should be on the dirtyIO scheduler. Elixir has a really really short timeout on NIF executions on the default scheduler. The "dirty IO" scheduler allows it to run on a background thread and not cause a NIF crash.

```rust
#[rustler::nif(schedule = "DirtyIo")]
```

host_core/native/hostcore_wasmcloud_native/src/lib.rs
```diff
@@ -70,10 +75,123 @@ rustler::init!(
...
+                claims: convert_provider_claims(claims),
+                target_bytes,
+                contract_id,
+                vendor,
+            },
+        ))
+    })
+}
+
+#[rustler::nif]
```
autodidaddict, line 157: This should also be on the dirty IO scheduler

host_core/native/hostcore_wasmcloud_native/src/par.rs
```diff
@@ -27,25 +27,30 @@ pub(crate) fn get_vendor(par: &ProviderArchive) -> Result<String, Error> {
...
-                ..Default::default()
-            })
-        }
+        Some(c) => Ok(convert_provider_claims(c)),
         None => Err(Error::Term(Box::new("No claims found in provider archive"))),
     }
 }
 
+/// Convert provider claims to our crate claims type
+pub fn convert_provider_claims(
```
autodidaddict, line 36: a) should this be public?
b) would this be more appropriate inside an `impl Into<crate::Claims> for wascap::jwt:Claims....` ?

autodidaddict (changes requested):
Some minor refactoring nits

host_core/native/hostcore_wasmcloud_native/src/par.rs
thomastaylor312: Oh sorry, this was left over from me importing the crate into my original bindle only crate. However, I think this is definitely better as a `From` impl, so I'll go add that

autodidaddict (approved):

PR #325, opened 2021-12-31
{WIP] Integrity poll of the lattice state from OTP

janitha09:
@autodidaddict 

The lattice observer builds a "lattice" that represents the state from the events that the observer observes. It makes sense to do an integrity scan - if an event was missed, The code is a simple way of polling the known good state from OTP. The state is constructed and maintained in a genserver within the wasmcloud_host  project

Is all the information below needed to reconstruct the lattice. I am thinking in terms of separating the runtime (dynamic data) from the model (static)

```
%WasmcloudHost.Lattice.StateMonitor.State{
  claims: %{
    "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5" => %{
      call_alias: "",
      caps: "wasmcloud:httpserver",
      iss: "ACOJJN6WUP4ODD75XEBKKTCCUJJCY5ZKQ56XVKYK4BEJWGVAOOQHZMCW",
      name: "echo",
      rev: "4",
      sub: "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5",
      tags: "",
      version: "0.3.2"
    },
    "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M" => %{
      call_alias: "",
      caps: "",
      iss: "ACOJJN6WUP4ODD75XEBKKTCCUJJCY5ZKQ56XVKYK4BEJWGVAOOQHZMCW",
      name: "httpserver",
      rev: "1631827501",
      sub: "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M",
      tags: "",
      version: "0.14.4"
    }
  },
  hosts: %{
    "NAYEXIY6HDQ2RO6UO5V5O7P6PAVH6WGTC5QPHHWEVV2UX4WSAHVNCHVJ" => %{
      actors: %{
        "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5" => %{
          count: 2,
          status: "Healthy"
        }
      },
      labels: %{
        "hostcore.arch" => "x86_64",
        "hostcore.os" => "linux",
        "hostcore.osfamily" => "unix"
      },
      providers: %{
        {"VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M", "default"} => %{
          contract_id: "wasmcloud:httpserver",
          status: "Healthy"
        }
      }
    }
  },
  linkdefs: %{
    {"MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5",
     "wasmcloud:contract",
     "default"} => %{
      provider_key: "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M",
      values: %{}
    }
  },
  refmaps: %{
    "wasmcloud.azurecr.io/echo:0.3.2" => "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5",
    "wasmcloud.azurecr.io/httpserver:0.14.4" => "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M"
  }
}
```

How is the GenServer data being persisted? 
What is the authoritative source of truth for the lattice (lattice observer or OTP)?
In the event that the system crashes (host_core) what is the recovery mechanism?


autodidaddict:
the OTP host shouldn't be used to send any state to wadm. `wadm` has its own NATS subscriber that listens on the event stream and then applies those events in-memory to a lattice observer. 

The state monitor that's in the `wasmcloud_host` repository is deprecated and will eventually be replaced with a NATS subscriber that uses a lattice-observer to generate state on the fly from the event stream.

In other words, when things are "complete" then both wasmcloud_host and wadm will be using the same lattice-observer library to generate state from an event stream. They will not communicate with each other to compare notes on state because they're maintaining state for two entirely different reasons.

In wadm, the state is being persisted in a Redis cache and the wadm application is designed to run as a BEAM cluster with automatic failover and high availability.

autodidaddict:
I do like the idea of an integrity check, but wadm is an opt-in part of the ecosystem.. a wasmCloud host can and does run without wadm available on the network for such an integrity check. I think one avenue I'd like to explore here (and I think I've got an issue around here somewhere for it) would be to allow people to configure the event stream in JetStream as durable and then allow replays from that stream to reconstitute state afresh.

autodidaddict:
> What is the authoritative source of truth for the lattice (lattice observer or OTP)?

There isn't one. There are different authorities depending on what you're doing. The wasmcloud_host's view of the lattice state is purely one of convenience for users of the "washboard". If the host goes down and comes back up, it's fine if it's lost observed state. It'll refresh its host list through the next sequence of heartbeats.

> In the event that the system crashes (host_core) what is the recovery mechanism?

For the wasmcloud_host process, it's fine. The main function of the host is to run the wasmCloud plumbing, it doesn't actually need to maintain the runtime lattice state so long as it gets replay of the "lattice cache", which is the raw data like link definitions and image refs and claims. wasmcloud_host's state monitor is a convenience. Ultimately that's going to be replaced by a use of the lattice monitor, but it won't change the ephemeral nature of this cache.

As mentioned, `wadm` is the process that has a highly available state cache because it's required for managing active deployments whereas the "washboard"'s state is a handy convenience.

PR #326, opened 2021-12-31
feat(*) - enable hot reloading in docker

tadejstanic:
After reading this issue - https://github.com/wasmCloud/wasmcloud-dev-site/issues/83 I decided to test if adding `inotify-tools to the `wasmcloud` docker image would enable hot reloading through docker container. It turned out that by adding this package the wasmcloud Dashboard is able to get filesystem events from the mounted volume. 

Tested on OSx, not sure how if this works on Windows machines too, so if there is anybody that can test this before it gets merged would be great.


autodidaddict (dismissed):

autodidaddict:
@tadejstanic please fix the DCO line item (commit with signature) then this should be fine.

brooksmtownsend (dismissed):

brooksmtownsend (approved):

PR #330, opened 2022-01-08
Host pings now trigger heartbeats, health status only emitted when it changes

autodidaddict:
This PR accomplishes the following:
* Host pings now function like proper radar and will emit a heartbeat as soon as they receive a ping
* Health checks are now only emitted when the state changes, which keeps the event log from being spammed by endless healthy status events when running at large-scale steady state.

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -63,7 +63,7 @@ defmodule HostCore.Actors.ActorModule do
   end
 
   def halt(pid) do
-    GenServer.call(pid, :halt_and_cleanup)
+    if Process.alive?(pid), do: GenServer.call(pid, :halt_and_cleanup)
   end
 
   def health_check(pid) do
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -25,7 +25,8 @@ defmodule HostCore.Actors.ActorModule do
       :invocation,
       :claims,
       :subscription,
-      :ociref
+      :ociref,
+      :healthy
     ]
   end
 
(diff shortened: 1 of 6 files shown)


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -62,7 +63,7 @@ defmodule HostCore.Actors.ActorModule do
   end
 
   def halt(pid) do
-    GenServer.call(pid, :halt_and_cleanup)
+    if Process.alive?(pid), do: GenServer.call(pid, :halt_and_cleanup)
   end
```
brooksmtownsend, line 67: I wonder if this is a place that we can handle actors panicking, by knowing that it's not alive when we call `halt`

Just a thought, no change requested

host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -184,10 +187,25 @@ defmodule HostCore.Providers.ProviderModule do
...
 
-    case res do
-      {:ok, _body} -> publish_health_passed(state)
-      {:error, _} -> publish_health_failed(state)
-    end
+    # Only publish health check pass/fail when state changes
+    state =
+      case res do
+        {:ok, _body} ->
+          if !state.healthy do
```
brooksmtownsend, line 194: nit: this case statement could read easier with snippets like 
```
{:ok, _body} when !state.healthy -> ...
```

not strongly opposed to the current style

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -468,12 +466,10 @@ defmodule HostCore.WebAssembly.Imports do
   end
 
   defp guest_request(_api_type, context, agent, op_ptr, ptr) do
-    Logger.debug("Guest request")
```
brooksmtownsend, line 471: It's logs like this that make me wish Logger had a trace that we could be this verbose in, but I digress

brooksmtownsend (commented):
A couple comments with an optional change, just re-request review and I'll drop a ✅

host_core/lib/host_core/providers/provider_module.ex
autodidaddict: fun fact, you can't use `!` in a guard clause.

brooksmtownsend: Oh, `when not`, that's neat

brooksmtownsend (approved):

PR #340, opened 2022-01-27
Ensured tests purge before running

brooksmtownsend:
Tests that depend on or establish linkdefs, claims, or OCI references need to purge the stream before they run their test. The event watcher is a means-to-an-end here, e.g. it isn't strictly necessary for these tests but the initialization function is useful. Happy to split this out into another piece of logic if needed.

### DIFF
--- a/host_core/lib/host_core/providers/provider_module.ex
+++ b/host_core/lib/host_core/providers/provider_module.ex
@@ -90,7 +90,7 @@ defmodule HostCore.Providers.ProviderModule do
       publish_provider_oci_map(claims.public_key, link_name, oci)
     end
 
-    Process.send(self(), :do_health, [:noconnect, :nosuspend])
+    Process.send_after(self(), :do_health, 5_000)
     :timer.send_interval(@thirty_seconds, self(), :do_health)
 
     {:ok,
--- a/host_core/mix.exs
+++ b/host_core/mix.exs
@@ -10,6 +10,11 @@ defmodule HostCore.MixProject do
       elixir: "~> 1.12",
       start_permanent: Mix.env() == :prod,
       deps: deps(),
+			rustler_crates: [
+				hostcore_wasmcloud_native: [
+					mode: (if Mix.env() == :dev, do: :debug, else: :release)
+				]
+			],
       dialyzer: [plt_add_deps: :apps_direct]
     ]
   end
--- a/host_core/mix.exs
+++ b/host_core/mix.exs
@@ -10,11 +10,11 @@ defmodule HostCore.MixProject do
       elixir: "~> 1.12",
       start_permanent: Mix.env() == :prod,
       deps: deps(),
-			rustler_crates: [
-				hostcore_wasmcloud_native: [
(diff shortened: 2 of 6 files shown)


autodidaddict:
Does the starting of the event watcher automatically do a purge? Trying to find where the purge is happening in this PR

brooksmtownsend:
> Does the starting of the event watcher automatically do a purge? Trying to find where the purge is happening in this PR

Yep exactly, https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/test/support/event_watcher.exs#L21-L31

brooksmtownsend:
https://github.com/wasmCloud/wasmcloud-otp/runs/4969397684?check_suite_focus=true#step:8:1222-1232 seems to support my theory but this still hasn't solved it 🤔

autodidaddict (dismissed):

brooksmtownsend:
I see this:
```
Starting capability provider VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M instance 11889b11-e2d6-4b71-aaea-b0bbd1df18d6 with nats url 127.0.0.1:4222
 INFO wasmcloud_provider_httpserver > httpserver starting listener on 127.0.0.1:8083 for actor MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5
 INFO wasmcloud_provider_httpserver > httpserver starting listener on 127.0.0.1:8081 for actor MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E
 INFO wasmcloud_provider_httpserver > httpserver starting listener on 127.0.0.1:8082 for actor MBW3UGAIONCX3RIDDUGDCQIRGBQQOWS643CVICQ5EZ7SWNQPZLZTSQKU
 INFO wasmcloud_provider_httpserver > httpserver starting listener on 127.0.0.1:8084 for actor MCRACEYKJLCP7NLVMUTO2SCQ26JFL3Z4TBIXPXDE7KPBI2GEH5DA57L5
```

Which implies the stream is not properly purged. Investigating

autodidaddict:
host purge doesn't actually purge the whole stream.. it's just the in-host cache (right?)

autodidaddict:
> > Does the starting of the event watcher automatically do a purge? Trying to find where the purge is happening in this PR
> 
> Yep exactly, https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/test/support/event_watcher.exs#L21-L31
> 
> edit: actually, I bet this happens once per test-module and not once per test. The `init` is called when we call `start_link` edit2: Nope I was correct, `setup` is called once per-test thinking

This is purging the JetStream data .. but does it empty the Host's linkdefs and claims cache?

brooksmtownsend:
> This is purging the JetStream data .. but does it empty the Host's linkdefs and claims cache?

Oh man, I don't think it does

autodidaddict (dismissed):

autodidaddict (approved):

PR #346, opened 2022-02-16
Initial implementation of support for supplemental config via service

autodidaddict:
At startup, if the `WASMCLOUD_CONFIG_SERVICE` environment variable is truthy,  the wasmCloud host will attempt to pull supplemental configuration from `wasmbus.cfg.{latttice}`.

This configuration may contain a list of actors to auto-start, a list of providers to auto-start, and a map of registry credentials to be used for pulling from secure OCI/bindle registries.

Not included in this PR (to be added later): support for receiving _push_ updates from a config service as a notification that registry credentials have changed.

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -144,7 +144,7 @@ defmodule HostCore.Host do
         if !Map.has_key?(prov, "imageReference") || !Map.has_key?(prov, "linkName") do
           Logger.error("Not enough information on auto-start provider configuration. Bypassing.")
         else
-          if String.starts_with?(prov["imageReference"], "bindle") do
+          if String.starts_with?(prov["imageReference"], "bindle://") do
             HostCore.Providers.ProviderSupervisor.start_provider_from_bindle(
               prov["imageReference"],
               prov["linkName"]
@@ -160,7 +160,7 @@ defmodule HostCore.Host do
 
       as
       |> Enum.each(fn actor ->
-        if String.starts_with?(actor, "bindle") do
+        if String.starts_with?(actor, "bindle://") do
           HostCore.Actors.ActorSupervisor.start_actor_from_bindle(actor)
         else
           HostCore.Actors.ActorSupervisor.start_actor_from_oci(actor)
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -131,15 +131,15 @@ defmodule HostCore.Host do
   # supp config's prestarts
   @impl true
   def handle_continue(:process_supp_config, %State{supplemental_config: sc} = state) do
-    aps = Map.get(sc, "autoStartProviders", [])
-    as = Map.get(sc, "autoStartActors", [])
+    autostart_providers = Map.get(sc, "autoStartProviders", [])
+    autostart_actors = Map.get(sc, "autoStartActors", [])
 
     Logger.info(
-      "Processing supplemental configuration: #{length(aps)} providers, #{length(as)} actors"
+      "Processing supplemental configuration: #{length(autostart_providers)} providers, #{length(autostart_actors)} actors"
     )
 
     Task.start(fn ->
-      aps
+      autostart_providers
       |> Enum.each(fn prov ->
         if !Map.has_key?(prov, "imageReference") || !Map.has_key?(prov, "linkName") do
           Logger.error("Not enough information on auto-start provider configuration. Bypassing.")
@@ -158,7 +158,7 @@ defmodule HostCore.Host do
         end
       end)
 
-      as
+      autostart_actors
       |> Enum.each(fn actor ->
         if String.starts_with?(actor, "bindle://") do
           HostCore.Actors.ActorSupervisor.start_actor_from_bindle(actor)
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -9,7 +9,7 @@ defmodule HostCore.Host do
 
   defmodule State do
     @moduledoc false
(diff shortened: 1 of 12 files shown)


host_core/lib/host_core/host.ex
```diff
@@ -73,13 +73,102 @@ defmodule HostCore.Host do
...
+    if config_service_enabled?(opts) do
+      {:ok, state, {:continue, :load_supp_config}}
+    else
+      {:ok, state}
+    end
+  end
+
+  # truthy
+  defp config_service_enabled?(opts) do
+    String.upcase(opts.config_service_enabled) in [
```
brooksmtownsend, line 92: so many suggestions for this but in the interest in keeping this function under 100 lines I'll hold back

host_core/lib/host_core/host.ex
```diff
@@ -73,13 +73,102 @@ defmodule HostCore.Host do
...
+      "Processing supplemental configuration: #{length(aps)} providers, #{length(as)} actors"
+    )
+
+    Task.start(fn ->
+      aps
+      |> Enum.each(fn prov ->
+        if !Map.has_key?(prov, "imageReference") || !Map.has_key?(prov, "linkName") do
+          Logger.error("Not enough information on auto-start provider configuration. Bypassing.")
+        else
+          if String.starts_with?(prov["imageReference"], "bindle") do
```
brooksmtownsend, line 147: to prevent a wild edge case where someone has their OCI server named `bindlebasket.azurecr.io`, we should probably check for starting with `bindle://`

host_core/lib/host_core/host.ex
```diff
@@ -73,13 +73,102 @@ defmodule HostCore.Host do
...
+              prov["imageReference"],
+              prov["linkName"]
+            )
+          end
+        end
+      end)
+
+      as
+      |> Enum.each(fn actor ->
+        if String.starts_with?(actor, "bindle") do
```
brooksmtownsend, line 163: same comment re: bindle

host_core/native/hostcore_wasmcloud_native/src/client.rs
```diff
@@ -43,30 +43,67 @@ impl TokenManager for PickYourAuth {
...
+
+// If the bindle ID is in "secure registry" form, just take the invoice portion of invoice@server
+pub(crate) fn normalize_bindle_id(bindle_id: &str) -> String {
+    let parts: Vec<_> = bindle_id.split('@').collect();
+    if parts.len() == 2 {
+        parts[0].to_owned()
+    } else {
+        bindle_id.to_owned()
+    }
+}
```
brooksmtownsend, line 109: wont this function get called with the reference `bindle://invoice@server`?

host_core/native/hostcore_wasmcloud_native/src/client.rs
```diff
@@ -43,30 +43,67 @@ impl TokenManager for PickYourAuth {
...
     } else {
-        PickYourAuth::None(NoToken)
-    };
+        match (
+            var(BINDLE_PASSWORD_ENV),
+            var(BINDLE_USER_NAME_ENV),
+            var(BINDLE_TOKEN_ENV),
+        ) {
+            (Ok(pw), Ok(username), _) => PickYourAuth::Http(HttpBasic::new(&username, &pw)),
+            (_, _, Ok(token)) => PickYourAuth::LongLived(LongLivedToken::new(&token)),
```
brooksmtownsend, line 60: After looking at https://github.com/deislabs/bindle/blob/main/docs/README.md#configuring-authentication, does Bindle even support just token authentication? I see github oauth2 and http basic auth. cc @thomastaylor312

brooksmtownsend (changes requested):

host_core/native/hostcore_wasmcloud_native/src/client.rs
autodidaddict: No, the `bindle://` prefix is stripped off at the Elixir level before the bindle client stuff gets invoked via NIF.

host_core/native/hostcore_wasmcloud_native/src/client.rs
autodidaddict: the `LongLivedToken` struct actually comes from the bindle client so ... I'm assuming it's supported? This is mostly code that existed prior to this, I just refactored the nested if statements into a match.

thomastaylor312: Yeah, there is a token thing. Right now the authn/z features in bindle are half baked. I have an idea behind making that better but I still need to write it up

brooksmtownsend (dismissed):

host_core/lib/host_core/host.ex
```diff
@@ -73,13 +73,102 @@ defmodule HostCore.Host do
...
+    {:noreply, state}
+  end
+
+  # NOTE - we do this work in a handle continue because we need the registry credentials
+  # to be a part of this process's state prior to attempting to start the components in the
+  # supp config's prestarts
+  @impl true
+  def handle_continue(:process_supp_config, %State{supplemental_config: sc} = state) do
+    aps = Map.get(sc, "autoStartProviders", [])
+    as = Map.get(sc, "autoStartActors", [])
```
connorsmith256, line 135: Nit: these names aren't clear to me and don't seem to use the same abbreviation pattern. Consider using a few more characters to disambiguate

connorsmith256 (dismissed):

host_core/lib/host_core/host.ex
autodidaddict: Good point

PR #347, opened 2022-02-17
Fix/provider archive memory spike

brooksmtownsend:
@thomastaylor312 wrote the vast majority of this PR, I only stepped in to deal with inconsistent tests.

This PR upgrades the wasmcloud NIF to use a newer version of `provider-archive` which more efficiently unpacks provider archives for memory constrained devices.

The TL;DR regarding inconsistent tests is that there was a race condition between a provider receiving hostdata with link definitions and actually subscribing to the linkdef topic so we were intermittently losing link definitions, causing HTTP requests and keyvalue invocations to fail

### DIFF
--- a/host_core/native/hostcore_wasmcloud_native/src/lib.rs
+++ b/host_core/native/hostcore_wasmcloud_native/src/lib.rs
@@ -179,27 +179,29 @@ fn get_provider_bindle(
 
         let claims: Claims = claims.into();
 
-        // Now get the parcel
-        let mut written = 0usize;
-        let mut file = get_provider_file(&par::cache_path(
+        // Now get the parcel (if it doesn't already exist on disk)
+        if let Some(mut file) = get_provider_file(&par::cache_path(
             &claims.public_key,
             claims.revision.unwrap_or_default(),
             &contract_id,
             &link_name,
         )?)
-        .await?;
-        let mut stream = bindle_client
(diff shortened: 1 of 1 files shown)


autodidaddict (approved):

thomastaylor312:
Ok, this has been updated with the latest provider-archive version and should be ready for final review

autodidaddict (approved):

host_core/native/hostcore_wasmcloud_native/src/par.rs
```diff
@@ -40,18 +40,18 @@ pub(crate) fn extract_target_bytes(par: &ProviderArchive) -> Result<Vec<u8>, Err
 }
 
 pub(crate) fn cache_path(
-    subject: String,
-    rev: u32,
-    contract_id: String,
-    link_name: String,
+    subject: &str,
+    rev: i32,
```
brooksmtownsend, line 44: I think this should still be `u32` unless there's a reason I'm missing to support negative numbers (feels weird for filepath)

autodidaddict: yeah revisions are never negative.

thomastaylor312: I had actually changed it, but the actual claims type from wascap is an `i32`, so it didn't really make sense to change it. I just changed it to match what things were set to

autodidaddict (approved):

PR #350, opened 2022-02-22
Replaced webpack with esbuild

### LINKED ISSUE #251: [SPIKE] NPM dependencies for wasmcloud_host
We rely on node/NPM in order to build the dependencies we need for the wasmcloud_host dashboard. We're currently using Node 14.X, which is not the latest version. We also generate lockfiles in version 2, which always generates a little warning since node 14 (or whatever dependency it is) expects version 1. Lastly, there are some `npm audit` warnings about outdated/vulnerable packages, we should update what we can while remaining functional with CoreUI.

As a complete alternative, Phoenix 1.60 (which is now used as of #250) allows for esbuild and ditching `npm` entirely. If we can get this to work with CoreUI, that would be a great alternative.

brooksmtownsend:
Fixes #251 

Spent some fun weekend time replacing the webpack config with esbuild, and got it working! This removes the node/npm and python dependencies from wasmcloud.

Draft PR as I'm still working on fully testing this.


.github/workflows/release.yml
```diff
@@ -42,6 +43,30 @@ jobs:
...
+        run: |
+          mix local.rebar --force 
+          mix local.hex --force
+          mix deps.get
+
+      - name: Compile Phoenix Assets
+        working-directory: ${{env.working-directory}}
+        # esbuild + dart sass are only included in the dev runtime environment
+        env:
+          MIX_ENV: dev
```
thomastaylor312, line 66: Just to double check, doing this with `dev` isn't going to screw up `prod` stuff down the line?

thomastaylor312 (approved):
Mmmmm...that freshly cleaned code smell

.github/workflows/release.yml
brooksmtownsend: That's right, we just only include the SASS compiler in the development environment to keep the prod env as minimal as possible

brooksmtownsend: Yeah I used the localized `env` option which should only apply to that one step

thomastaylor312: Yes, got that too. Obviously I am not explaining myself well 😅 . 

So the concern here is: Will debug/dev on these steps end up including large or unneeded resources in the final release tarball (even though the rest are using `prod`)?

brooksmtownsend: Ah okay I think I understand better. So this step does end up doing a bit of unnecessary computation by compiling dependencies for dev, but the only assets that will carry over to the release tarball are the minified CSS files that are compiled. Put another way, running this step in dev mode is just a tool to keep the sass dependency out of the release binary and keep it only in the dev environment.

brooksmtownsend: Apologies for merging before this is resolved

PR #354, opened 2022-02-24
Initial commit of invocation chunking support

autodidaddict:

### DIFF
--- a/.github/workflows/wasmcloud_host.yml
+++ b/.github/workflows/wasmcloud_host.yml
@@ -52,36 +52,36 @@ jobs:
           mix local.hex --force
           mix do deps.get, deps.compile
 
-      - name: Check Formatting
-        working-directory: ${{env.working-directory}}
-        run: mix format --check-formatted
+      # - name: Check Formatting
+      #   working-directory: ${{env.working-directory}}
+      #   run: mix format --check-formatted
 
-      - name: Run Credo
-        working-directory: ${{env.working-directory}}
-        continue-on-error: true # Don't fail entire action with refactoring opportunities for now
-        run: mix credo --strict
+      # - name: Run Credo
+      #   working-directory: ${{env.working-directory}}
+      #   continue-on-error: true # Don't fail entire action with refactoring opportunities for now
+      #   run: mix credo --strict
 
-      - name: Run Tests
-        working-directory: ${{env.working-directory}}
-        run: |
-          WASMCLOUD_LATTICE_PREFIX=$(echo "${{ runner.os }}__${{ matrix.otp }}__${{ matrix.elixir }}__${{ env.working-directory }}" | sed 's/\./_/g') \
-          mix test
+      # - name: Run Tests
+      #   working-directory: ${{env.working-directory}}
+      #   run: |
+      #     WASMCLOUD_LATTICE_PREFIX=$(echo "${{ runner.os }}__${{ matrix.otp }}__${{ matrix.elixir }}__${{ env.working-directory }}" | sed 's/\./_/g') \
+      #     mix test
 
-      - name: Retrieve PLT Cache
-        uses: actions/cache@v2
-        id: plt-cache
-        with:
-          path: wasmcloud_host/priv/plts
-          key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-plts-${{ hashFiles('wasmcloud_host/mix.exs', 'wasmcloud_host/mix.lock') }}
+      # - name: Retrieve PLT Cache
+      #   uses: actions/cache@v2
+      #   id: plt-cache
+      #   with:
+      #     path: wasmcloud_host/priv/plts
+      #     key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-plts-${{ hashFiles('wasmcloud_host/mix.exs', 'wasmcloud_host/mix.lock') }}
 
-      - name: Create PLTs
-        working-directory: ${{env.working-directory}}
-        if: steps.plt-cache.outputs.cache-hit != 'true'
-        run: |
-          mkdir -p priv/plts
-          mix dialyzer --plt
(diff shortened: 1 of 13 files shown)


autodidaddict:
Especially interested in your thoughts @stevelr as this has direct impact on `wasmbus_rpc`

host_core/native/hostcore_wasmcloud_native/src/lib.rs
```diff
@@ -17,10 +21,17 @@ use wascap::prelude::*;
...
+mod objstore;
 mod oci;
 mod par;
 mod task;
 
+lazy_static! {
+    static ref CHUNKING_STORE: RwLock<Option<ObjectStore>> = RwLock::new(None);
+}
+
+const CHONKY_THRESHOLD_BYTES: usize = 1024 * 700; // 500KB
```
connorsmith256, line 33: Supernit: 700KB?

host_core/native/hostcore_wasmcloud_native/src/lib.rs
```diff
@@ -104,6 +118,29 @@ rustler::init!(
...
+    let jsoptions = if let Some(d) = config.get("js_domain").cloned() {
+        JetStreamOptions::new().domain(&d)
+    } else {
+        JetStreamOptions::new()
+    };
+    let js = nats::jetstream::JetStream::new(nc, jsoptions);
+    let lattice = config
+        .get("lattice")
+        .cloned()
+        .unwrap_or("default".to_string());
```
connorsmith256, line 136: Does falling back actually help us recover here? Would it be better to return an error if there's no lattice prefix?

connorsmith256 (approved):

host_core/native/hostcore_wasmcloud_native/src/lib.rs
autodidaddict: image

host_core/native/hostcore_wasmcloud_native/src/lib.rs
autodidaddict: Nearly everything in the wasmCloud ecosystem uses `default` as the lattice when one isn't supplied, so it felt natural to make this behave that way as well.

connorsmith256 (approved):

host_core/lib/host_core/host.ex
```diff
@@ -371,14 +371,15 @@ defmodule HostCore.Host do
...
       case :ets.lookup(:config_table, :config) do
         [config: config_map] ->
           {"#{config_map[:prov_rpc_host]}:#{config_map[:prov_rpc_port]}",
-           config_map[:prov_rpc_jwt], config_map[:prov_rpc_seed], config_map[:prov_rpc_tls]}
+           config_map[:prov_rpc_jwt], config_map[:prov_rpc_seed], config_map[:prov_rpc_tls],
+           config_map[:rpc_timeout]}
 
         _ ->
-          {"127.0.0.1:4222", "", ""}
+          {"127.0.0.1:4222", "", "", 2000}
```
brooksmtownsend, line 382: aha so this is why my rpc timeout change didn't configure anything :P

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -409,13 +424,31 @@ defmodule HostCore.WebAssembly.Imports do
...
+    if Map.get(ir, "content_length", bsize) > bsize do
+      with {:ok, bytes} <- HostCore.WasmCloud.Native.dechunk_inv(invid) do
+        bytes
+      else
+        _ ->
+          ir["msg"]
+      end
+    else
+      ir["msg"]
+    end
```
brooksmtownsend, line 449: super nit, could push the if statement into with
```
with true <- Map.get(ir, "content_length, bsize) > bsize ,
        {:ok, bytes} <- ... do
           bytes
else
_ -> ir["msg"]
end
```

brooksmtownsend (changes requested):
Only real request is that I'd love to see a test with this working, is there a somewhat simple scenario that we could use to test this?

autodidaddict:
> Only real request is that I'd love to see a test with this working, is there a somewhat simple scenario that we could use to test this?

Yeah the tick is finding an actor I can use that has a contract that allows for a payload big enough to chunk

brooksmtownsend (approved):

host_core/mix.exs
```diff
@@ -33,15 +33,16 @@ defmodule HostCore.MixProject do
       {:wasmex, "~> 0.5.0"},
       {:msgpax, "~> 2.3"},
       {:rustler, "~> 0.22.0"},
+      {:jason, "~> 1.2.2"},
       {:gnat, "~> 1.2"},
       {:cloudevents, "~> 0.4.0"},
       {:uuid, "~> 1.1"},
-      {:jason, "~> 1.2.2"},
+
```
brooksmtownsend, line 40: Good for Jason, moving its way up the dependency ladder

autodidaddict: Everybody loves an underdog story

brooksmtownsend (approved):

.github/workflows/wasmcloud_host.yml
```diff
@@ -52,36 +52,36 @@ jobs:
...
+      #   run: mix format --check-formatted
 
-      - name: Run Credo
-        working-directory: ${{env.working-directory}}
-        continue-on-error: true # Don't fail entire action with refactoring opportunities for now
-        run: mix credo --strict
+      # - name: Run Credo
+      #   working-directory: ${{env.working-directory}}
+      #   continue-on-error: true # Don't fail entire action with refactoring opportunities for now
+      #   run: mix credo --strict
```
brooksmtownsend, line 62: We can keep these sections, just need to skip the `mix test` section

brooksmtownsend (changes requested):

brooksmtownsend (approved):
This time it'll work ™️

PR #355, opened 2022-02-25
Return an error to the host if it fails to read a par file

protochron:
`par_from_file` in the Rust NIF returns a generic argument error if it
fails to load. It can fail for a number of reasons, so instead of
returning a generic error, return the error message generated from the
underlying provider library.


autodidaddict (approved):
This is a good catch... `BadArg` will actually cause the calling process to crash with Elixir's `BadArgument` exception.

brooksmtownsend (approved):

PR #360, opened 2022-03-03
Terminate provider module when linked binary stops

### LINKED ISSUE #359: [BUG] Capability provider crash does not close ProviderModule
**Describe the bug**

When a capability provider closes itself without being asked to, due to a crash or to a graceful exit, the ProviderModule receives the DOWN message from port but does not exit. This leaves provider subscriptions open and gives the false pretense that the provider is still running as it will still show in host inventories and the module keeps running.

When a provider closes by sending a DOWN message, we should clean up subscriptions, memory, and close out any other resources.

autodidaddict:
Fixes #359


autodidaddict:
Tests run and pass locally 10/10 times. :crossed_fingers: hopefully GH will do the same

brooksmtownsend (approved):
LGTM, tested locally with a provider that panics and it's removed from the washboard and I see the stopped event on the event stream

PR #366, opened 2022-03-09
Actor module now safely terminates when Wasmex fails to start

autodidaddict:
There exists an edge case that results in a double-publish of actor start failed events (control server message handles a start call that results in a Wasmex start failure). However, this is acceptable considering the cost and impracticality of trying to suppress the double-publish. No known downstream consumers of this event would be negatively impacted by receiving it twice.

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -51,7 +51,11 @@ defmodule HostCore.Actors.ActorModule do
   end
 
   def claims(pid) do
-    GenServer.call(pid, :get_claims)
+    if Process.alive?(pid) do
+      GenServer.call(pid, :get_claims)
+    else
+      %{}
+    end
   end
 
   def ociref(pid) do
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -280,22 +294,31 @@ defmodule HostCore.Actors.ActorModule do
...
+        end)
 
-    Wasmex.start_link(%{bytes: bytes, imports: imports})
-    |> prepare_module(agent, oci)
+        publish_oci_map(oci, claims.public_key)
+        {:ok, agent}
+
+      {:error, e} ->
+        Logger.error("Failed to start actor: #{inspect(e)}")
+        HostCore.ControlInterface.Server.publish_actor_start_failed(claims.public_key, inspect(e))
```
brooksmtownsend, line 319: I don't love the double publish but I can't find a better place to centralize the "start failed" logic 🤔 perhaps we can revisit this in the case where Wasmex crashes happen for more than failed imports.

brooksmtownsend (approved):

host_core/lib/host_core/actors/actor_module.ex
autodidaddict: this fix actually covers all Wasmex failures, not just failed imports.

PR #368, opened 2022-03-15
Support for structured (JSON) logging

autodidaddict:
Adds new environment variables to toggle structured logging (default is off because unstructured is easier for humans to read) and at what level. In other words, if you're using structured logging you can override the log level with an environment variable.

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -173,7 +173,7 @@ defmodule HostCore.Host do
 
   @impl true
   def terminate(reason, state) do
-    Logger.debug("Host termination requested: #{inspect(reason)}")
+    Logger.debug("Host termination requested: #{inspect(reason)}", reason: reason)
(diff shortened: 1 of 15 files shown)


host_core/lib/host_core/host.ex
```diff
@@ -404,7 +404,9 @@ defmodule HostCore.Host do
       config_json: config_json,
       default_rpc_timeout_ms: timeout,
       cluster_issuers: cluster_issuers(),
-      invocation_seed: cluster_seed()
+      invocation_seed: cluster_seed(),
+      # In case providers want to be aware of this for their own logging
+      structured_logging_enabled: structured_logging_enabled
```
brooksmtownsend, line 409: very nice 👍🏻

brooksmtownsend (approved):
Looks great, the additional detail with the keyword lists on the logs are 👌🏻.

PR #373, opened 2022-03-17
fix(host_core): Avoids crashing when passing a JS domain

thomastaylor312:
The new chunk config was unintentionally adding a value to the original `config`
object rather than adding to the chunk config

### DIFF
--- a/host_core/lib/host_core.ex
+++ b/host_core/lib/host_core.ex
@@ -135,7 +135,7 @@ defmodule HostCore do
 
     chunk_config =
       if config.js_domain != nil do
-        Map.put(config, "js_domain", config.js_domain)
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core.ex
```diff
@@ -135,7 +135,7 @@ defmodule HostCore do
 
     chunk_config =
       if config.js_domain != nil do
-        Map.put(config, "js_domain", config.js_domain)
+        Map.put(chunk_config, "js_domain", config.js_domain)
```
autodidaddict, line 138: Freaking typos.

autodidaddict (approved):

thomastaylor312:
I know we are close to a 0.54. Did we just want to release this with 0.54 or should we create a release branch for 0.53 and release a patch?

Issue #369, opened 2022-03-15
[BUG] Dashboard values cannot handle base64 config


brooksmtownsend:
**Describe the bug**

wasmCloud providers generally support the `config_b64` link definition values, e.g. `config_64=aHR0cHM6Ly93d3cueW91dHViZS5jb20vd2F0Y2g/dj1kUXc0dzlXZ1hjUQ==`, however our link definition values are separated by an `=` delimiter in the UI.

Here is our logic for parsing link definitions: https://github.com/wasmCloud/wasmcloud-otp/blob/main/wasmcloud_host/lib/wasmcloud_host_web/live/components/define_link_component.ex#L32-L38. As you can see we are splitting the string by an `=` sign and chunking every 2, which will break for the trailing `=` symbols that commonly follow base64 encoded strings. Our logic in that component needs to be updated to handle this case.

emattiza:
I'll take a stab at this one given i worked on the same issue in [wasmcloud/wash#249](https://github.com/wasmCloud/wash/issues/249)

emattiza:
> I'll take a stab at this one given i worked on the same issue in [wasmcloud/wash#249](https://github.com/wasmCloud/wash/issues/249)

Free to close. Thanks team!

PR #377, opened 2022-04-08
[BUG] Dashboard values cannot handle base64 '=' padding characters

emattiza:
Addresses #369

Utilize the parts option to only split once to ensure correct key=value decoding behavior.

See [Hex docs](https://hexdocs.pm/elixir/String.html#split/3-options)

Matches what wash does. [See wash](https://github.com/wasmCloud/wash/blob/4e9ccc406b7ca996fb1e6153aa93732d62047186/src/util.rs#L133)

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host_web/live/components/define_link_component.ex
+++ b/wasmcloud_host/lib/wasmcloud_host_web/live/components/define_link_component.ex
@@ -32,7 +32,7 @@ defmodule DefineLinkComponent do
         value_list ->
           value_list
           |> String.split(",")
-          |> Enum.flat_map(fn s -> String.split(s, "=") end)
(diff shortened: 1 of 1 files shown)


autodidaddict (commented):
LGTM once the DCO check is fixed.

autodidaddict:
I love how simple this solution turned out to be.

emattiza:
> I love how simple this solution turned out to be.

There's still time to write it all with a parser combinator... 😈

autodidaddict:
> > I love how simple this solution turned out to be.
> 
> There's still time to write it all with a parser combinator... smiling_imp

Or better yet, create the "properly splits key=value pairs" programming language. It'll be huge.

autodidaddict (approved):

brooksmtownsend (approved):
These elixir folks thought of everything 😄  Thanks Evan!

Issue #331, opened 2022-01-10
Upgrade wasmex to latest version


brooksmtownsend:
According to [this commit](https://github.com/tessi/wasmex/commit/aeb1f7eeb41b74ebf8b88f218de55ae8a4f10d5d) and this [tweet](https://twitter.com/wasmerio/status/1480555668549558276?s=20) the newest version of wasmex is coming soon with support for Elixir 1.13. We should upgrade the host to use this version as well as our GH Actions CI to test against this version as soon as it's ready

brooksmtownsend:
closed by #378

PR #378, opened 2022-04-09
update wasmex,wasmer,rustler,& rust dependencies

stevelr:
Updated:

- [wasmex 0.7.0 from 0.5.0](https://github.com/tessi/wasmex/blob/main/CHANGELOG.md), which updates wasmer to 2.1.1
- [rustler 0.24 from 0.23](https://github.com/rusterlium/rustler/blob/master/CHANGELOG.md)
- updated other rust dependencies via 'cargo update'

Signed-off-by: Steve <steve@cosmonic.com>

host_core tests pass. Creating as draft pr to allow additional testing


stevelr:
also updated build runner from ubuntu 18.04 to 20.04.

Apparently, this version of ubuntu includes a mono dev environment that automatically runs a monodoc http server on port 8084, so had to change the echo/httpserver test to use a different port number.
The monodoc service can be stopped with 
`sudo systemctl stop mono-xsp4.service` but it was simpler to change the port than the workflow script.

stevelr:
I don't know how to get rid of those expected checks for 1.12, since 1.12 isn't in the matrix anymore.

thomastaylor312:
@stevelr We can remove those from branch protections once we merge this

thomastaylor312 (commented):
One minor change: Also, do we need to update the elixir version of the lattice controller as well so they play nice together?

host_core/mix.exs
```diff
@@ -30,11 +30,13 @@ defmodule HostCore.MixProject do
...
   defp deps do
     [
-      {:wasmex, "~> 0.5.0"},
+      {:wasmex, "~> 0.7.0"},
       {:msgpax, "~> 2.3"},
-      {:rustler, "~> 0.22.0"},
+      {:rustler, "~> 0.24.0"},
       {:jason, "~> 1.2.2"},
       {:gnat, "~> 1.2"},
+      # erlavro isn't used, but this version upgrades dependency of cloudevents 0.4.0 to use rebar3
```
autodidaddict, line 38: This feels a little off. We're including a dependency that we don't need in order to upgrade the way another dependency works?
surely there's a better way to do this?

autodidaddict (commented):
Just a question about adding a dependency we don't need...

autodidaddict:
> One minor change: Also, do we need to update the elixir version of the lattice controller as well so they play nice together?

If, by `lattice controller`, you mean `lattice observer`, then it doesn't need to change yet because there's no direct dependency. At the moment, the only coupling between the lattice observer and the OTP host is the payload contract on the `wasmbus.evt.*` topic.

autodidaddict:
This fixes #331

host_core/mix.exs
autodidaddict: There's got to be a better way to fix this, but for now let's just move forward.

autodidaddict (approved):

PR #380, opened 2022-04-14
fix(host_core): Makes chunking work in all places

thomastaylor312:
This now makes it so that actor to actor call chunking works as well
as provider to actor call chunking.

Note: I will come back and add a test for actor to actor chunking once
I merge in the new changes to the petclinic


autodidaddict (approved):
Chonk approved. 

image

brooksmtownsend (approved):
very nice. Chonky boi above is chonky but he looks a little sad :(

brooksmtownsend (changes requested):
@thomastaylor312 go ahead and bump to `0.54.2` so we can get this released 👍🏻

brooksmtownsend (approved):

PR #381, opened 2022-04-14
chore(*): Bumps version for patch release

thomastaylor312:


thomastaylor312:
NOTE: I purposefully left out bumping the helm chart because of the git diff issue where it won't detect the chart changes if we release at the same time as the tag. I will do that in a follow up PR after we cut the release tag

PR #383, opened 2022-04-15
Updates the host to support actors written in TinyGo

autodidaddict:
* invokes the `_start` function, which is exported by TinyGo's `main()` function
* exposes a fake WASI `fd_write` function because tinygo modules require this even if no code in the module utilizes it.


brooksmtownsend (approved):
👀

thomastaylor312 (approved):

PR #387, opened 2022-04-25
Adds support for per-instance annotaions on providers and actors

autodidaddict:
While this supports generic key-value pairs for annotations in the start instructions received from the control interface, its initial use will be to allow [wadm](https://github.com/wasmCloud/wadm) to tag individual instances with the application deployment to which it belongs.

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -14,9 +14,7 @@ defmodule HostCore.Actors.ActorSupervisor do
     DynamicSupervisor.init(strategy: :one_for_one)
   end
 
-  @spec start_actor(bytes :: binary(), oci :: String.t(), count :: Integer.t()) ::
-          {:error, any} | {:ok, [pid()]}
-  def start_actor(bytes, oci \\ "", count \\ 1) when is_binary(bytes) do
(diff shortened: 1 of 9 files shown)


host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -14,9 +14,7 @@ defmodule HostCore.Actors.ActorSupervisor do
     DynamicSupervisor.init(strategy: :one_for_one)
   end
 
-  @spec start_actor(bytes :: binary(), oci :: String.t(), count :: Integer.t()) ::
```
thomastaylor312, line 17: Nit: don't we still want this spec here?

thomastaylor312 (approved):

host_core/lib/host_core/actors/actor_supervisor.ex
autodidaddict: Eventually. I couldn't get vscode to validate the spec. I'll try and check it back in and see if it works without warnings

brooksmtownsend: I had trouble getting dialyxir to recognize it but I think I got warnings that `Integer.t()` and `Map.t()` aren't real types and should be `integer()` and `map()`

PR #391, opened 2022-05-09
Disambiguate error message for denied invocation

### LINKED ISSUE #364: [BUG] Disambiguate missing capability claim from missing link definition
**Describe the bug**
There are two scenarios that can lead to the same error message, missing capability claims and a missing or misconfigured link definition. This is confusing to the user as these are two totally separate problems and should be indicated separately. The message is in the form of:
```
Invocation not authorized: missing claim for <contract_id>
```

This is a fine error message when an actor is missing a capability claim, but not when the link definition is missing and therefore unable to continue. It's worth mentioning that the invocation is and should be denied for both of these scenarios, this bug report is simply to accurately report the reason in the logs.

**More Information**
here's the function https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/web_assembly/imports.ex#L244, which is called when authorized: false. that can happen either:
contract is not in capability claims https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/web_assembly/imports.ex#L236
call is not verified, e.g. link definition not found https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/web_assembly/imports.ex#L221

brooksmtownsend:
Fixes #364 

This PR disambiguates a previous confusing error message for a denied invocation into 2 parts
1. A missing link definition for a contract ID / link name pair (link not configured)
2. A missing capability claim (Actor not allowed to invoke that capability)

Using the kvcounter example, here's what the response error messages look like now. For a successful invocation:
```
> curl localhost:8080/sup
{"counter":1}%  
```

For a kvcounter actor that is missing the `wasmcloud:keyvalue` claim:
```
> curl localhost:8080/sup
{"error":"Host send error Invocation not authorized: missing capability claim for wasmcloud:keyvalue"}%
```

For a kvcounter actor that is either missing a link definition for the `wasmcloud:keyvalue` contract, or does not have a configured link definition for the `wasmcloud:keyvalue` contract on the `default` link name (these are similar, but different)
```
> curl localhost:8080/sup
{"error":"Host send error Invocation not authorized: missing link definition for wasmcloud:keyvalue on default"}%  
```

### DIFF
--- a/host_core/lib/host_core/web_assembly/imports.ex
+++ b/host_core/lib/host_core/web_assembly/imports.ex
@@ -253,9 +253,34 @@ defmodule HostCore.WebAssembly.Imports do
     %{token | authorized: true}
   end
 
+  # Deny invocation due to missing link definition for a contract ID and link name
+  defp invoke(
+         _token = %{
+           verified: false,
+           authorized: false,
+           agent: agent,
+           namespace: namespace,
+           prefix: prefix
+         }
+       ) do
+    Agent.update(agent, fn state ->
+      %State{
+        state
+        | host_error:
+            "Invocation not authorized: missing link definition for #{namespace} on #{prefix}"
+      }
+    end)
+
+    0
+  end
+
+  # Deny invocation due to missing capability claim
   defp invoke(_token = %{authorized: false, agent: agent, namespace: namespace}) do
     Agent.update(agent, fn state ->
-      %State{state | host_error: "Invocation not authorized: missing claim for #{namespace}"}
+      %State{
+        state
+        | host_error: "Invocation not authorized: missing capability claim for #{namespace}"
+      }
     end)
 
     0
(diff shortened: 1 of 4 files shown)


host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -253,9 +253,34 @@ defmodule HostCore.WebAssembly.Imports do
     %{token | authorized: true}
   end
 
+  # Deny invocation due to missing link definition for a contract ID and link name
+  defp invoke(
+         _token = %{
+           verified: false,
+           authorized: false,
```
connorsmith256, line 260: Nit: do we need this `authorized` check? I noticed `authorized` will always be false if `verified` is false

connorsmith256 (approved):

autodidaddict (dismissed):

host_core/lib/host_core/web_assembly/imports.ex
brooksmtownsend: Fair, we will deny it if the link isn't verified

connorsmith256 (approved):

PR #396, opened 2022-05-11
fixed release env in host core

brooksmtownsend:
Same thing as #389 just for host_core


.github/workflows/release.yml
```diff
@@ -61,6 +61,7 @@ jobs:
 
       - name: Compile Phoenix Assets
         working-directory: ${{env.working-directory}}
+        shell: bash
```
connorsmith256, line 64: Is this because the default shell for windows is powershell?

connorsmith256 (approved):

.github/workflows/release.yml
brooksmtownsend: Yep that's correct

PR #406, opened 2022-05-25
Propagating selected env vars to providers

autodidaddict:
Rather than add the environment variables to the `env_values` field on the host info struct, I decided to propagate the environment variables as actual env vars. This means that capability providers need to do no work in order to handle the environment variables and there won't be any race conditions where a provider might query an env var before it extracts the host info struct.

Initially this is to enable open telemetry in the capability providers but we can also add select variables to the propagation list later


brooksmtownsend (approved):
In the future, for specific use cases, I think this sets us up well to include a list of "special" environment variables to propagate in the host config

PR #408, opened 2022-05-28
Add IPv6 support for NATS connections

protochron:
Add a new environment variable (`WASMCLOUD_ENABLE_IPV6`) that enables
IPv6 support for establishing NATS connections. Without this value, the
Gnat client cannot connect to IPv6-only endpoints.

In theory we could just always enable it, since the underlying `:inet6`
option is dual-stack, but wasmcloud hosts may end up running on systems
that do not support IPv6.

Signed-off-by: Dan Norris <protochron@users.noreply.github.com>


autodidaddict (approved):

protochron:
@brooksmtownsend Other than some weird embedded device, I'd be surprised if _any_ system doesn't support IPv6 unless it's specifically disabled in the kernel (famous last words, I know). I figured flagging it on and off was the safer option, but we can default it to on later if we find that it's not actually a problem.

PR #410, opened 2022-06-01
fix(host_core): Makes sure the structured_logging_enabled is a bool

thomastaylor312:
If you set the environment variable, the value would be serialized as
a string, which made Rust not very happy (and probably any other
strongly typed language). This is a small fix that ensures that it is
an actual bool before it is encoded into the config map

### DIFF
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -423,10 +423,14 @@ defmodule HostCore.Host do
       cluster_issuers: cluster_issuers(),
       invocation_seed: cluster_seed(),
       # In case providers want to be aware of this for their own logging
-      structured_logging_enabled: structured_logging_enabled
+      structured_logging_enabled: to_bool(structured_logging_enabled)
     }
     |> Jason.encode!()
   end
+
+  defp to_bool(val) when is_binary(val), do: String.downcase(val) == "true"
+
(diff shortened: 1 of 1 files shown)


host_core/lib/host_core/host.ex
```diff
@@ -423,11 +423,15 @@ defmodule HostCore.Host do
...
       # In case providers want to be aware of this for their own logging
-      structured_logging_enabled: structured_logging_enabled
+      structured_logging_enabled: to_bool(structured_logging_enabled)
     }
     |> Jason.encode!()
   end
 
+  defp to_bool(val) when is_binary(val), do: String.downcase(val) == "true"
+
+  defp to_bool(_), do: false
```
connorsmith256, line 433: I don't think this case is currently possible -- since the config value defaults to [false](https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/config_plan.ex#L55) -- but if we ever change the default to `true`, will `to_bool` still return `false`?

brooksmtownsend: Agreed with Connor here, if it's a boolean we can just default to the boolean

thomastaylor312: Oh good catch. I'll add that case

connorsmith256 (approved):

PR #411, opened 2022-06-06
fix(host_core): Makes sure trace id is propagated on every request

thomastaylor312:
This also solves an issue with a health check having a sticky trace ID
from the previous request


autodidaddict (approved):

brooksmtownsend (approved):
Just to be clear, even if you don't have an OTEL exporter running this works fine?

LGTM

PR #415, opened 2022-06-13
Add extended information to host ping reply and heartbeats

### LINKED ISSUE #414: [FEATURE] Suppress Heartbeat Logging
**Is your feature request related to a problem? Please describe.**
Would it be possible to suppress or minimize the heartbeat log without suppressing other debug logging.

```
16:19:23.152 [debug] Publishing heartbeat
```

**Describe the solution you'd like**
Maybe just log change in heartbeat status

**Additional context**
This log makes up a high percentage of my logging ATM and I would just like to minimize if possible

autodidaddict:
This also fixes #414


host_core/lib/host_core/control_interface/server.ex
```diff
@@ -50,10 +50,36 @@ defmodule HostCore.ControlInterface.Server do
...
+            {nil, "localhost", "localhost", "localhost", "default"}
+        end
+
       res = %{
         id: HostCore.Host.host_key(),
+        labels: HostCore.Host.host_labels(),
         friendly_name: HostCore.Host.friendly_name(),
-        uptime_seconds: div(total, 1000)
+        uptime_seconds: ut_seconds,
+        uptime_human: ut_human,
```
connorsmith256, line 75: Thanks for including both formats 🎉

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -50,10 +50,36 @@ defmodule HostCore.ControlInterface.Server do
...
+      ut_human =
+        ut_seconds
+        |> Timex.Duration.from_seconds()
+        |> Timex.Format.Duration.Formatters.Humanized.format()
+
+      {js_domain, ctl_host, prov_rpc_host, rpc_host, lattice_prefix} =
+        case :ets.lookup(:config_table, :config) do
+          [config: config_map] ->
+            {config_map.js_domain, config_map.ctl_host, config_map.prov_rpc_host,
+             config_map.rpc_host, config_map.lattice_prefix}
```
connorsmith256, line 64: Isn't it possible for some of these to be set, but not all? We should have default/fallback values per field

connorsmith256 (changes requested):

host_core/lib/host_core/control_interface/server.ex
autodidaddict: The sanitization step ("post process config") ensures that there's always a meaningful value/default for these. So the only time we need to return those other defaults is in the (should never occur) case of config not existing in the ETS table.

connorsmith256 (approved):

brooksmtownsend (approved):

PR #418, opened 2022-06-14
send log messages to stderr

connorsmith256:


host_core/lib/host_core.ex
```diff
@@ -20,7 +20,10 @@ defmodule HostCore do
     if config.enable_structured_logging do
       :logger.add_handler(:structured_logger, :logger_std_h, %{
         formatter: {HostCore.StructuredLogger.FormatterJson, []},
-        level: config.structured_log_level
+        level: config.structured_log_level,
+        config: %{
+          type: :standard_error
```
autodidaddict, line 25: Is it really `type` in code and `device` in the config file?

connorsmith256: It's `type` for `logger_std_h`, and `device` for `Logger`:

https://www.erlang.org/doc/man/logger_std_h.html

https://github.com/elixir-lang/elixir/blob/main/lib/logger/lib/logger/backends/console.ex#L25

It was difficult to find documentation for these 😞

autodidaddict: I love consistent inconsistency

autodidaddict (approved):

PR #420, opened 2022-06-15
Refactor to consumer supervisors for RPC subscriptions

autodidaddict:
Included in this PR:
* Refactor of manual RPC subscriptions for actors into consumer supervisors
  * Subscriptions are now reused per actor and not per instance 
* Misc error handling and edge case enhancements
* Health checks for actors are no longer performed (corresponding events no longer emitted)

### DIFF
--- a/host_core/config/config.exs
+++ b/host_core/config/config.exs
@@ -21,6 +21,14 @@ config :opentelemetry,
   span_processor: :batch,
   traces_exporter: :none
 
+# Subscription retention is a test-enabling tool. When enabled, a subscriber for
+# an actor _will not unsubscribe_ after all instances of that actor have been
+# removed from a host. This helps in testing because the constant churning of
+# stopping, unsubscribing, starting, and re-subscribing in the middle of a test
+# suite causes failure and race conditions on a large scale.
+#
+# tl;dr - leave subscription retention off at all times unless you're running
+# tests.
 config :host_core,
   retain_rpc_subscriptions: false
 
--- a/host_core/config/config.exs
+++ b/host_core/config/config.exs
@@ -21,4 +21,7 @@ config :opentelemetry,
   span_processor: :batch,
   traces_exporter: :none
 
+config :host_core,
+  retain_rpc_subscriptions: false
+
 import_config "#{config_env()}.exs"
--- a/host_core/lib/host_core/actors/actor_rpc_server.ex
+++ b/host_core/lib/host_core/actors/actor_rpc_server.ex
@@ -11,9 +11,8 @@ defmodule HostCore.Actors.ActorRpcServer do
       ) do
     pk = topic |> String.split(".") |> Enum.at(-1)
 
-    # Randomly choose a running instance to handle this request.
-    # optimization for later - in the future, we might be able to detect an instance that isn't currently
-    # handling a request (isn't busy) to optimize this
+    # Choose the least busy (smallest message queue length) actor in the registry
+    # as the target for the inbound RPC
     #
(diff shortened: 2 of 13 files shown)


host_core/config/config.exs
```diff
@@ -21,4 +21,7 @@ config :opentelemetry,
   span_processor: :batch,
   traces_exporter: :none
 
+config :host_core,
+  retain_rpc_subscriptions: false
```
thomastaylor312, line 25: Wouldn't we want the default to be `true` or am I misreading/misinterpreting the code?

thomastaylor312 (approved):
One minor question, but otherwise LGTM

host_core/config/config.exs
autodidaddict: default should be `false`. You only ever want to do this retention in test. I should probably document the ramifications of this in the config file(s)

thomastaylor312: Yeah, documenting that sounds like a good idea

host_core/lib/host_core/actors/actor_rpc_server.ex
```diff
@@ -0,0 +1,44 @@
...
+          body: _body,
+          reply_to: _reply_to,
+          topic: topic
+        } = msg
+      ) do
+    pk = topic |> String.split(".") |> Enum.at(-1)
+
+    # Randomly choose a running instance to handle this request.
+    # optimization for later - in the future, we might be able to detect an instance that isn't currently
+    # handling a request (isn't busy) to optimize this
```
brooksmtownsend, line 16: This sounds like a great optimization to make later 👍🏻

host_core/lib/host_core/actors/actor_rpc_supervisor.ex
```diff
@@ -0,0 +1,62 @@
...
+    prefix = HostCore.Host.lattice_prefix()
+    topic = "wasmbus.rpc.#{prefix}.#{claims.public_key}"
+
+    cs_settings = %{
+      connection_name: :lattice_nats,
+      module: HostCore.Actors.ActorRpcServer,
+      subscription_topics: [
+        %{topic: topic, queue_group: topic}
+      ]
+    }
```
brooksmtownsend, line 35: Interesting, so instead of each instance subscribing to a queue group now it's only one per-host, so the queue groups will now round robin between hosts (effectively)

host_core/lib/host_core/host.ex
```diff
@@ -67,6 +67,12 @@ defmodule HostCore.Host do
       Logger.warn(warning)
     end
 
+    if Application.get_env(:host_core, :retain_rpc_subscriptions, false) do
+      Logger.warn(
+        "RPC Subscription Retention Enabled. This host instance is NOT stable in production!"
+      )
+    end
```
brooksmtownsend, line 74: Is this for a future update? I'm a little confused for the reasoning behind retaining an RPC subscription

brooksmtownsend: nvm, saw latest commit

host_core/lib/host_core/actors/actor_rpc_supervisor.ex
autodidaddict: yes

thomastaylor312 (approved):

brooksmtownsend (approved):

autodidaddict:
For posterity, this is where Jose mentions that getting process info externally locks the target: https://elixirforum.com/t/is-erlang-process-info-pid-message-queue-len-a-heavy-operation/20358/7?u=lpil

connorsmith256 (approved):

PR #422, opened 2022-06-21
Use round robin to dispatch RPC to actor instances

autodidaddict:

### DIFF
--- a/host_core/lib/host_core/actors/actor_rpc_server.ex
+++ b/host_core/lib/host_core/actors/actor_rpc_server.ex
@@ -1,5 +1,6 @@
 defmodule HostCore.Actors.ActorRpcServer do
   require Logger
+  alias HostCore.Actors.CallCounter
   use Gnat.Server
 
   def request(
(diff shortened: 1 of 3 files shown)


host_core/lib/host_core/actors/actor_rpc_server.ex
```diff
@@ -11,29 +12,17 @@ defmodule HostCore.Actors.ActorRpcServer do
       ) do
     pk = topic |> String.split(".") |> Enum.at(-1)
 
-    # NOTE - dispatch doesn't invoke the handler if no registry entries exist for the given key
+    case Registry.lookup(Registry.ActorRegistry, pk) do
+      [] ->
+        {:error, "Actor #{pk} is not running on this host. RPC call skipped."}
```
brooksmtownsend, line 17: Interesting edge case here. The scenario that causes this, a host that's listening on an RPC topic for an actor that's no longer running, is puzzling.

Perhaps here we could have the ActorRpcServer terminate that specific subscription? That may be a bad pattern, but if this message is received and there's no actor to handle it we probably shouldn't allow more messages to come to this host

autodidaddict: The RPC server will terminate the subscription when the actor is all out of instances. This is just to protect against ultra-low-chance race conditions where the actor process has died before the subscription has finished cleaning up.

autodidaddict: It's one of those things that should never realistically happen, but because code-wise it's physically possible, I still wanted to account for it to keep another process from crashing.

host_core/lib/host_core/actors/actor_rpc_server.ex
```diff
@@ -11,29 +12,17 @@ defmodule HostCore.Actors.ActorRpcServer do
...
-    #   {pid, _ql} =
-    #     entries
-    #     |> Enum.map(fn {pid, _value} ->
-    #       {pid, Process.info(pid, :message_queue_len) |> elem(1)}
-    #     end)
-    #     |> Enum.sort(&(elem(&1, 1) <= elem(&2, 1)))
-    #     |> Enum.at(0)
+      actors ->
+        next_index = CallCounter.read_and_increment(pk)
+        {pid, _value} = Enum.at(actors, rem(next_index, length(actors)))
```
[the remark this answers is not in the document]
brooksmtownsend: We'll just want to document that wasmcloud actors are designed to run for at max 97 years before you'll need to rotate your host. We can get around this by providing a free wasmCloud ops runbook to the grandchildren of any developers though.

PR #424, opened 2022-06-22
Bumped chart to 0.55

brooksmtownsend:
Seeking an opinion from @thomastaylor312, do we want to include some of the things from the otel docker compose with this chart?

If not, this should be good to go once 0.55 is released

### DIFF
--- a/wasmcloud_host/chart/templates/deployment.yaml
+++ b/wasmcloud_host/chart/templates/deployment.yaml
@@ -91,6 +91,10 @@ spec:
             - name: WASMCLOUD_JS_DOMAIN
               value: {{ .Values.wasmcloud.config.jetstreamDomain | quote }}
             {{- end }}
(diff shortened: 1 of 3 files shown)


wasmcloud_host/chart/templates/deployment.yaml
```diff
@@ -91,6 +91,10 @@ spec:
             - name: WASMCLOUD_JS_DOMAIN
               value: {{ .Values.wasmcloud.config.jetstreamDomain | quote }}
             {{- end }}
+            {{- if .Values.wasmcloud.config.registry.url }}
+            - name: OCI_REGISTRY
+              value: {{ .Values.wasmcloud.config.registry.url | quote }}
+            {{- end }}
             {{- if .Values.wasmcloud.config.registry.username }}
```
thomastaylor312, line 98: Just need to add in the new structured logging and OTEL variables here too

PR #425, opened 2022-06-22
Renamed structured logging field

brooksmtownsend:
This is a change in the interest of backwards compatibility and deserializing the hostdata struct


connorsmith256 (approved):
Do we want to leave the env var named as-is? I'm very mildly concerned about using different names, for searchability reasons

brooksmtownsend:
> Do we want to leave the env var named as-is? I'm very mildly concerned about using different names, for searchability reasons

@connorsmith256 I think in the interest of backwards compatibility that would be nice, no changes required, but I could be convinced

brooksmtownsend:
Renamed to `enable_structured_logging` per @autodidaddict 's request, cc @stevelr

autodidaddict (approved):

PR #428, opened 2022-06-28
Actor termination candidates now filtered by optional annotations

autodidaddict:
If an annotation map is supplied to the stop actor control interface function (and thereby the `terminate_actor` function), those annotations will be used to narrow the pool of candidate actors against which the stop operation works.

For context, `wadm` will supply such annotations where one annotation contains the "appspec" label, a key that wadm uses to identify which resources belong to which applications.

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -285,6 +285,7 @@ defmodule HostCore.Actors.ActorSupervisor do
       Registry.lookup(Registry.ActorRegistry, public_key)
       |> Enum.filter(fn {pid, _v} ->
         existing = HostCore.Actors.ActorModule.annotations(pid)
+        # Property of maps - map a is contained within b if b.merge(a) == b
         Map.merge(existing, annotations) == existing
       end)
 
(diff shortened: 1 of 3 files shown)


host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -279,8 +280,14 @@ defmodule HostCore.Actors.ActorSupervisor do
...
 
   # Terminate `count` instances of an actor
-  def terminate_actor(public_key, count) when count > 0 do
-    actors = Registry.lookup(Registry.ActorRegistry, public_key)
+  def terminate_actor(public_key, count, annotations) when count > 0 do
+    actors =
+      Registry.lookup(Registry.ActorRegistry, public_key)
+      |> Enum.filter(fn {pid, _v} ->
+        existing = HostCore.Actors.ActorModule.annotations(pid)
+        Map.merge(existing, annotations) == existing
```
brooksmtownsend, line 288: This might be worth a comment for anyone who's not aware of the merge case here :)

host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -296,8 +303,12 @@ defmodule HostCore.Actors.ActorSupervisor do
   end
 
   # Terminate all instances of an actor
-  def terminate_actor(public_key, 0) do
+  def terminate_actor(public_key, 0, annotations) do
     Registry.lookup(Registry.ActorRegistry, public_key)
+    |> Enum.filter(fn {pid, _v} ->
+      existing = HostCore.Actors.ActorModule.annotations(pid)
+      Map.merge(existing, annotations) == existing
```
brooksmtownsend, line 310: Same comment re: commenting on merge

brooksmtownsend (approved):
LGTM, minor nit

PR #429, opened 2022-06-29
gracefully restart an actor's RPC supervisor

connorsmith256:
Fix for https://github.com/wasmCloud/wasmcloud-otp/issues/430

### DIFF
--- a/host_core/config/config.exs
+++ b/host_core/config/config.exs
@@ -27,15 +27,4 @@ config :opentelemetry,
   span_processor: :batch,
   traces_exporter: :none
 
-# Subscription retention is a test-enabling tool. When enabled, a subscriber for
-# an actor _will not unsubscribe_ after all instances of that actor have been
-# removed from a host. This helps in testing because the constant churning of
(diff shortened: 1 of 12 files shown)


brooksmtownsend (commented):
With your testing, did this end up falling over in any way if you start an actor, stop an actor, and then attempt to start 100 actors simultaneously?

For best results, I'd love to know what happens if you do this with an actor from a file just to remove any time spent to download OCI bytes

autodidaddict (changes requested):
Per discussion we decided that we wanted to modify `terminate_actor` such that it would both terminate the process _and_ delete the child spec from the parent.

brooksmtownsend:
Once this is released, let's cut 0.55.1 to fix the bug and also update the instructions in wasmcloud-dev-site to point to this version

autodidaddict (dismissed):

brooksmtownsend (dismissed):

autodidaddict (approved):

autodidaddict (approved):

PR #440, opened 2022-08-09
Changing Actor RPC from cast to call, adding rudimentary performance tests with Benchee

brooksmtownsend:
[Benchee](https://github.com/bencheeorg/benchee) is a benchmarking framework for Elixir applications. This PR introduces Benchee as a test dependency and a single test that calls the echo actor repeatedly as fast as possible for 100 milliseconds, both via NATS and by directly invoking the `request` method. These benchmark tests can serve both as a performance indicator for the wasmCloud runtime and a protection against future regressions.

Besides, what's better than tangible results? Here's some results from benchmarking, and improving, the actor RPC server with `GenServer.cast` and `GenServer.call`.

All of the tests below use the same basic variables, 1 second to "warmup" and 5 seconds to measure function invocation times. `nats_echo_request` uses the safe_req method to send a NATS message (which is routed locally) to handle an invocation, and `direct_echo_request` directly invokes the actor RPC handler without the network "hop". These were done on the following machine, which Benchee quotes as being down to the microsecond accuracy (linux is down to the nanosecond)
```
Operating System: macOS
CPU Information: Apple M1 Pro
Number of Available Cores: 10
Available memory: 16 GB
Elixir 1.13.4
Erlang 24.3.3

Benchmark suite executing with the following configuration:
warmup: 1 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 12 s
```

You can run any of these by doing (structured logging not required, just surpressing output):
```
WASMCLOUD_STRUCTURED_LOGGING_ENABLED=true WASMCLOUD_STRUCTURED_LOG_LEVEL=warn MIX_ENV=test mix test test/host_core/benchmark_test.exs
```

# Cast
Cast is used to send an **asynchronous** request to a server, in the actor RPC case this means an ActorModule will handle an invocation asynchronously and send a reply back once it's done handling the invocation. You see that the direct request seems to complete quicker, but the nats request measures when the _reply_ is received, not when the operation is accepted. This is a far more useful metric as it's how we invoke wasmCloud actors.

(for below, `ips` stands for instructions per second)

1 actor, 1 parallel request (note, this timed out and panicked since the actor didn't respond within 5 seconds to be terminated)
```
Name                          ips        average  deviation         median         99th %
direct_echo_request      672.99 K      0.00000 s  ±1387.21%      0.00000 s      0.00000 s
nats_echo_request       0.00050 K         2.00 s     ±0.02%         2.00 s         2.00 s

Comparison: 
direct_echo_request      672.99 K
nats_echo_request       0.00050 K - 1346690.11x slower +2.00 s
```
1 actor, 10 parallel requests (note, this timed out and panicked since the actor didn't respond within 5 seconds to be terminated)
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        7.03 K      0.00014 s   ±120.94%      0.00012 s      0.00056 s
nats_echo_request       0.00050 K         2.00 s     ±0.01%         2.00 s         2.00 s

Comparison: 
direct_echo_request        7.03 K
nats_echo_request       0.00050 K - 14061.36x slower +2.00 s
```
100 actors, 1 parallel request (note, this timed out and panicked since the actor didn't respond within 5 seconds to be terminated)
```
Name                          ips        average  deviation         median         99th %
direct_echo_request       43.24 K      0.00002 s   ±371.35%      0.00001 s      0.00014 s
nats_echo_request       0.00050 K         2.00 s     ±0.03%         2.00 s         2.00 s

Comparison: 
direct_echo_request       43.24 K
nats_echo_request       0.00050 K - 86596.55x slower +2.00 s
```
100 actors, 10 parallel requests (note, this timed out and panicked since the actor didn't respond within 5 seconds to be terminated)
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        5.41 K      0.00018 s   ±287.25%      0.00011 s      0.00235 s
nats_echo_request       0.00050 K         2.00 s     ±0.03%         2.00 s         2.00 s

Comparison: 
direct_echo_request        5.41 K
nats_echo_request       0.00050 K - 10833.37x slower +2.00 s
```

In fact, with the current `cast` setup, I was only able to send 100ms of constant, non-parallel requests and still finish the test without a timeout or the test crashing. This is the result of one of those tests
```
Name                          ips        average  deviation         median         99th %
direct_echo_request       42.01 K      0.00002 s   ±287.62%      0.00001 s      0.00013 s
nats_echo_request       0.00094 K         1.07 s     ±0.00%         1.07 s         1.07 s

Comparison: 
direct_echo_request       42.01 K
nats_echo_request       0.00094 K - 44909.86x slower +1.07 s
```

# Call
Call is used to send a **synchronous** request to a server, in the actor RPC case this means an ActorModule will handle an invocation and return the response immediately. You'll see that the direct request and the NATS request tend to be similar speeds, with the NATS request coming in slightly slower each time.

1 actor, 1 parallel request
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        1.41 K      710.46 μs    ±28.02%      683.38 μs     1118.54 μs
nats_echo_request          1.02 K      983.98 μs    ±33.28%      902.92 μs     2206.06 μs

Comparison: 
direct_echo_request        1.41 K
nats_echo_request          1.02 K - 1.38x slower +273.51 μs
```
1 actor, 10 parallel requests
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        136.91        7.30 ms    ±16.04%        6.91 ms       12.16 ms
nats_echo_request          129.51        7.72 ms    ±10.05%        7.50 ms       10.55 ms

Comparison: 
direct_echo_request        136.91
nats_echo_request          129.51 - 1.06x slower +0.42 ms
```
100 actors, 1 parallel request
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        1.25 K        0.80 ms    ±33.41%        0.76 ms        1.60 ms
nats_echo_request          0.91 K        1.10 ms    ±42.73%        0.98 ms        3.24 ms

Comparison: 
direct_echo_request        1.25 K
nats_echo_request          0.91 K - 1.38x slower +0.30 ms
```
100 actors, 10 parallel requests
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        802.81        1.25 ms    ±34.22%        1.17 ms        2.73 ms
nats_echo_request          609.23        1.64 ms    ±19.44%        1.59 ms        2.62 ms

Comparison: 
direct_echo_request        802.81
nats_echo_request          609.23 - 1.32x slower +0.40 ms
```

These numbers were chosen out of a hat and I'd be happy to stress the numbers even more. My overall conclusion and assumption here is that, because the Gnat.Server behavior already sends messages in an asynchronous task, combining that with the `GenServer.cast` results in too many concurrent requests hitting the Actor process inbox at one time, and the actors are actually far better suited to handle those requests _synchronously_ since the request receiver is non-blocking. I'm not 100% sure how this will impact performance across all scenarios, but one thing that I know for sure is using `cast`s too quickly can overwhelm actor modules (thanks Elixir for being so dang fast 😠)

Generally I'm seeing slowdowns as the benchmark attempts to handle more and more concurrent requests, but those are greatly improved by scaling actors.


autodidaddict:
This is amazing stuff! Thanks so much for all the diligence here!

autodidaddict (approved):

Issue #439, opened 2022-08-01
[RFC] Implement pluggable policy authority over NATS


autodidaddict:
# Summary

This is an RFC to request discussion on a feature idea regarding the implementation of a pluggable policy authority. The current version of the OTP wasmCloud host enforces security policy to prevent actors from communicating with capability providers for which they do not have claims, but that is the extent of the enforcement. There is no enforcement on actor-to-actor calls and any actor or capability provider can be started. 

# Rationale

In managed, compliance-controlled, multi-tenant, and many other environments, operators may want the ability to enforce more strict policy on the contents of each wasmCloud host. There may be scenarios specific to an organization where they may want to block one actor from communicating with other actors (or ban actor-to-actor calling altogether). Operators may want to be able to enforce a policy on which actors or providers can be started.

In today's system there is also no way to immediately flag a malicious actor to block it from being able to continue operating. If a policy server were to be consulted, then bad actors (or providers 😎) could be banned from performing any action.

# Desired Implementation

In the proposed implementation, if the wasmCloud host is started with the environment variable `WASMCLOUD_POLICY_TOPIC` set, then additional enforcement checks will be made when requests to start actors, start providers, or invoke actors and providers occur. When this topic is set, the following payload will be sent out on the topic:

```json
{
    "requestId": "string",
    "source": {
        "public_key": "Mxxx or Vxxx",
        "contract_id": "nullable",
        "link_name": "nullable",
        "capabilities": [ "wasmcloud:xxx", "wasmcloud:yyy" ],
        "issuer": "Axxx",
        "issued_on": "xxxx",
        "expires_in_mins": 32 
    },
    "target": {
        "public_key": "Mxxxx",
        "issuer": "Axxxx",
        "contract_id": "nullable",
        "link_name": "default"
    },
    "action": "start | call",
    "host": {
         "public_key": "Nxxxx",
         "labels": [
             "(key)" : "value"
         ],
        "valid_cluster_issuers": [ "Cxxxx", "Cxxxy" ],
     }
}
```

In response, the host will expect a result that has the following shape:

```json
{
    "actionPermitted": "true | false",
    "message": "error ipsum",
    "requestId": "string"
}
```

This response indicates whether the requested action is permitted or not. If the action is not permitted, a `message` field may optionally be supplied to provide contextual information _suitable for log emission_ to explain the policy failure. The `requestId` field is returned matching the corresponding request as a convenience.

[...]

PR #442, opened 2022-08-10
Evaluate operations against a policy enforcer over NATS

brooksmtownsend:
Implementation of the host-side of the RFC #439 

This PR includes a new module, `HostCore.Policy.Manager` which is responsible for taking a source, target, and action and querying a policy topic for an assertive allow or deny. This only occurs when a host is configured to use a policy topic, `WASMCLOUD_POLICY_TOPIC`, and quickly returns an allow in the case that no policy topic is configured.

In the case that a policy query times out (e.g. isn't answered by a policy enforcer) the wasmCloud host fails **closed** and will **not allow** the request to go through. Note that this means that a host configured with a policy topic that is invalid, or when a policy server goes down, no actors or providers can be started that haven't been previously authorized and no invocations can go through that haven't already happened between existing resources (cc @stevelr @autodidaddict)

When policy requests are denied due to invalid structures or timing out, those results will not be cached

### Some side effects that might not be initially obvious and need some extra thought
1. If a policy topic is configured on accident or with a typo, the first request to do any new action (start a new actor, perform an invocation, etc) will take 2 seconds longer as it waits to time out. This timeout probably needs to be adjusted as 2 seconds is enough to fail any invocation with the default RPC timeout (assuming it's coming from a provider)
2. If a policy request is invalid (e.g. missing an issuer key), it's allowed without consulting the policy engine. We control writing these requests in the host so it _shouldn't_ be an issue, but if an actor is missing an issuer I could see the argument to deny it.

### Still TODO:
- [x] Implement `WASMCLOUD_POLICY_CHANGES_TOPIC` to invalidate portions of the cache
- [x] Implement policy evaluation for `start_provider` and `perform_invocation`


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -405,7 +406,66 @@ defmodule HostCore.Actors.ActorModule do
...
+           issued_on: source_claims[:iat],
+           expires_in_mins: source_claims[:exp]
+         },
+         %{
+           public_key: target["public_key"],
+           contract_id: target["contract_id"],
+           link_name: target["link_name"],
+           issuer: target_claims[:iss]
+         },
+         "perform_invocation"
```
autodidaddict, line 434: we should probably put these action type constants into module consts

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -405,7 +406,66 @@ defmodule HostCore.Actors.ActorModule do
...
+    {:error, "actor is missing capability claims"}
+  end
+
+  # Deny invocation if policy enforcer does not permit it
+  defp perform_invocation(
+         {{_agent, true}, %{permitted: false} = policy_res},
+         _operation,
+         _payload
+       ) do
+    message = Map.get(policy_res, :message, "reason not specified")
```
autodidaddict, line 458: so many opportunities to provide a hilarious default message here.

autodidaddict: `Ye have been weighed, you have been measured, and ye have been found wanting`

host_core/lib/host_core/policy/manager.ex
```diff
@@ -0,0 +1,160 @@
+defmodule HostCore.Policy.Manager do
+  @moduledoc false
+  require Logger
+  # TODO: Gnat.server? Listen for
```
autodidaddict, line 4: does this comment still apply?

autodidaddict (commented):
looking good so far

host_core/lib/host_core/policy/manager.ex
```diff
@@ -0,0 +1,160 @@
...
+         expires_in_mins: _expires
+       }) do
+    :ok
+  end
+
+  defp validate_source(source) when is_map(source) do
+    # Narrow down missing fields by removing present fields from the list
+    missing_fields =
+      [:public_key, :capabilities, :issuer, :issued_on, :expires_in_mins]
+      |> Enum.reject(fn required_field -> Map.get(source, required_field) != nil end)
```
connorsmith256, line 107: `reject` with `!= nil` hurts my brain. Why not `filter` with `== nil`?

host_core/lib/host_core/policy/manager.ex
brooksmtownsend: nope! removed

host_core/lib/host_core/policy/manager.ex
brooksmtownsend: good call!

host_core/test/host_core/policy_test.exs
```diff
@@ -0,0 +1,213 @@
...
+      )
+
+    # Ensure the host doesn't start the actor that's denied
+    assert !(HostCore.Actors.ActorSupervisor.all_actors()
+             |> Map.keys()
+             |> Enum.any?(fn public_key ->
+               public_key == "MB2ZQB6ROOMAYBO4ZCTFYWN7YIVBWA3MTKZYAQKJMTIHE2ELLRW2E3ZW"
+             end))
+
+    # # Ensure the host doesn't start the provider that's denied
```
connorsmith256, line 207: Supernit: double comment

connorsmith256 (approved):

brooksmtownsend:
> @brooksmtownsend do you know why the build is failing checks?

Yep looks like I missed a refactor point. pushed a fix 🤞🏻

connorsmith256 (approved):

autodidaddict (approved):

PR #443, opened 2022-08-15
fix helm

Taction:
Fix the typo error in the template.


brooksmtownsend (approved):
Great catch @Taction! Thanks for submitting the PR fix 😄

PR #445, opened 2022-08-16
fix(web): fix dropdown duplicated ids

rovast:
image


brooksmtownsend (approved):
No more console errors! 🚀 Great catch and fix @rovast 
image

PR #446, opened 2022-08-17
bump host version to v0.56.0

connorsmith256:

### DIFF
--- a/wasmcloud_host/chart/Chart.yaml
+++ b/wasmcloud_host/chart/Chart.yaml
@@ -15,6 +15,6 @@ icon: https://github.com/wasmCloud/wasmcloud.com-dev/raw/main/static/images/wasm
 
 type: application
 
-version: 0.6.1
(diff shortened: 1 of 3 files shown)


brooksmtownsend (approved):

wasmcloud_host/chart/Chart.yaml
```diff
@@ -15,6 +15,6 @@ icon: https://github.com/wasmCloud/wasmcloud.com-dev/raw/main/static/images/wasm
 
 type: application
 
-version: 0.6.1
+version: 0.6.2
```
thomastaylor312, line 18: So there is a weird bug with the chart publisher stuff that it won't detect changes from the previous version if run when a tag is cut. Can you please manually trigger a chart build before tagging?

connorsmith256: Will do!

Issue #417, opened 2022-06-13
[BUG] Dashboard doesn't accept connections on `127.0.0.1`


brooksmtownsend:
**Describe the bug**

If you try to access the dashboard from `127.0.0.1`, the websocket spins forever and fails to authenticate. The "correct" URL is `localhost:4000`, but since these are functionally the same it would be great to accept either for the local dev purpose.

**To Reproduce**

1. Launch `wasmCloud` locally
2. Try to visit `http://127.0.0.1:4000` 
3. see washboard spin

PR #450, opened 2022-08-23
fix dashboard doesn't complain for every connection

Taction:
fix: #417.
Do not check the origin instead of checking it as `127.0.0.1`.

### DIFF
--- a/wasmcloud_host/config/prod.exs
+++ b/wasmcloud_host/config/prod.exs
@@ -13,6 +13,7 @@ config :wasmcloud_host, WasmcloudHostWeb.Endpoint,
   http: [:inet6, port: System.get_env("PORT") || 4000],
   # This is critical for ensuring web-sockets properly authorize.
   url: [host: "localhost", port: System.get_env("PORT")],
+  check_origin: false,
   cache_static_manifest: "priv/static/cache_manifest.json",
   server: true,
   root: ".",
--- a/wasmcloud_host/config/prod.exs
+++ b/wasmcloud_host/config/prod.exs
@@ -15,7 +15,7 @@ config :wasmcloud_host, WasmcloudHostWeb.Endpoint,
   url: [host: "localhost", port: System.get_env("PORT")],
   check_origin: [
     "//localhost:#{System.get_env("PORT") || 4000}",
-    System.get_env("ORIGIN") || "//127.0.0.1:#{System.get_env("PORT") || 4000}"
+    "//127.0.0.1:#{System.get_env("PORT") || 4000}"
   ],
   cache_static_manifest: "priv/static/cache_manifest.json",
   server: true,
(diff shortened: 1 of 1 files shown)


wasmcloud_host/config/prod.exs
```diff
@@ -13,6 +13,7 @@ config :wasmcloud_host, WasmcloudHostWeb.Endpoint,
   http: [:inet6, port: System.get_env("PORT") || 4000],
   # This is critical for ensuring web-sockets properly authorize.
   url: [host: "localhost", port: System.get_env("PORT")],
+  check_origin: false,
```
brooksmtownsend, line 16: It makes sense why this would fix the issue, but I don't know enough at first glance if this would cause security issues. What kinds of vulnerabilities do we introduce by not checking the origin of a request to the dashboard? (cc @jordan-rash @autodidaddict )

autodidaddict: the goal behind the way it was designed was to never allow remote access to this dashboard. So long as the default host on which it runs is one that is non-routable I'm okay with not checking origin

brooksmtownsend: What we can do is allow localhost & 127.0.0.1 unless it's overridden by the environment variable

brooksmtownsend (commented):
Unfortunately, at this level, environment variables are built into the release and not parsed at runtime. We do already have a dashboard host variable that overrides localhost, https://github.com/wasmCloud/wasmcloud-otp/blob/main/wasmcloud_host/lib/wasmcloud_host_web/endpoint.ex#L54.

brooksmtownsend:
@Taction can you check installing from `main` after #457 ? I believe I may have fixed this issue

Taction:
@brooksmtownsend Sure, I'll have a try, just to clarify I need to set WASMCLOUD_DASHBOARD_HOST and WASMCLOUD_DASHBOARD_ENV to enable my dashboard URL.

brooksmtownsend:
> @brooksmtownsend Sure, I'll have a try, just to clarify I need to set WASMCLOUD_DASHBOARD_HOST and WASMCLOUD_DASHBOARD_ENV to enable my dashboard URL.

I'm not sure what `WASMCLOUD_DASHBOARD_ENV` does, is that something you have locally?

Taction:
> I'm not sure what `WASMCLOUD_DASHBOARD_ENV` does, is that something you have locally?

Oh sorry, a typo mistake, it is `WASMCLOUD_DASHBOARD_PORT` actually.

brooksmtownsend:
@Taction I did actually fix a bug around parsing the port... you shouldn't have to specify anything now (version 0.57.1 of wasmCloud) if you want to access at localhost:4000 or 127.0.0.1:4000

brooksmtownsend:
Closing this as it's possible to fix the dashboard complaining after #457 , please feel free to re-open if the issue persists.

PR #447, opened 2022-08-19
Cloudevents lib update

brooksmtownsend:
this PR updates the cloudevents library which removes avro bindings (and the need for a cc for `snappyer`, a library we don't even use) to simplify the build process.

I also fixed a small bug in the makefile found in #444

PR #451, opened 2022-08-23
Mix releases, aarch64 host_core tarball, better Makefiles

brooksmtownsend:
Upgrading from distillery to mix releases alongside upgrading to OTP 25 due to failures in builds. Distillery is largely unmaintained at this point anyways and mix releases are the officially supported way to create releases.

This PR:
1. Changes wasmcloud_host and host_core application tarballs to be `mix releases` rather than Distillery releases. This is the officially supported method of creating application releases for Elixir applications and is much better maintained. This is also a breaking change as mix releases do not have the same command arguments.
2. Adds a host_core docker image that builds off of an alpine base. This is extracted as `aarch64-linux-musl-core.tar.gz`
3. Extracts the tarball from the `wasmcloud_host` arm64 Docker image and tags it as `aarch64-linux.tar.gz`
4. Updates our `mix.lock` dependencies, superceding #447 and simplifying dependencies
5. Cleans up Makefiles, notably adding `make build-image` as an easy, one-step target in both host_core and wasmcloud_host

Here's an example of the full, successful release, you can see tarballs for all of the above targets https://github.com/wasmCloud/wasmcloud-otp/actions/runs/2919947109

I've also done some spiking on BEAM burritos (work in https://github.com/wasmCloud/wasmcloud-otp/tree/feat/mmmm-beam-burrito if you're curious) however I'm thinking that this would require too much change. The speedbumps I hit with BEAM burritos are around:
1. Adding `zig` as a build tool dependency
7. Running a BEAM burrito when permissions are restricted (you have to extract the tarball somewhere)
8. Maintenance (the team behind burrito is fantastic, I'm simply concerned that issues in burrito releases may take longer than `mix release`s)

That being said, I'd still love to explore it in the future.


thomastaylor312 (approved):

.github/workflows/host_core.yml
```diff
@@ -16,17 +16,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, windows-2019, macos-10.15]
-        elixir: [1.13.3]
-        otp: [23, 24]
+        elixir: [1.13.4]
+        otp: [25]
```
autodidaddict, line 20: :tada:

autodidaddict (approved):

PR #454, opened 2022-08-31
Changed host_core release to copy musl nif

brooksmtownsend:
Got the action correct in #451 , but copied the wrong NIF

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -197,7 +197,7 @@ jobs:
       - uses: actions/download-artifact@v3
         with:
           path: ./host_core/priv/built/aarch64
-          name: aarch64-unknown-linux-gnu
+          name: aarch64-unknown-linux-musl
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -192,12 +193,12 @@ jobs:
       - uses: actions/download-artifact@v3
         with:
           path: ./host_core/priv/built/x86_64
-          name: x86_64-unknown-linux-gnu
+          name: x86_64-unknown-linux-musl
```
thomastaylor312, line 196: Should this still be gnu?

brooksmtownsend: No, this is just used for the docker image which is an alpine based image

brooksmtownsend: The mix release for x86 host_core still will use the gnu

thomastaylor312 (approved):

PR #456, opened 2022-09-01
Return error when attempting to overwrite existing linkdef

### LINKED ISSUE #371: [BUG] Link definitions should reject requests to overwrite existing definitions
**Describe the bug**

Link definitions are keyed in the form of `{actor, contract_id, link_name}`, as actors are only allowed to have one link definition per-contract ID and link name. Because this is not based on provider IDs two separate link definitions for different providers that share a contract ID and link name will overwrite the linkdef in the host cache _and_ be ignored by the capability provider.

**Solution**
If we attempt to [put a link definition](https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/linkdefs/manager.ex#L35) and the key already exists, we should instead refuse to cache that link definition. This should return an error to the control interface client indicating a link already exists and that the user should instead delete the existing linkdef

drowzy:
Use [`:ets.insert_new/2`](https://www.erlang.org/doc/man/ets.html#insert_new-2) to only insert link defs which does not already exist in the ets table. Closes #371

### DIFF
--- a/host_core/lib/host_core/control_interface/server.ex
+++ b/host_core/lib/host_core/control_interface/server.ex
@@ -158,8 +158,11 @@ defmodule HostCore.ControlInterface.Server do
              ) do
         {:reply, success_ack()}
       else
-        {:error, {:duplicate_key, {_, _, link_name}}} ->
-          {:reply, failure_ack("Duplicate link definition entry for '#{link_name}'")}
+        {:error, {:duplicate_key, {actor_id, contract_id, link_name}}} ->
+          {:reply,
+           failure_ack(
+             "Link definition for #{actor_id} on contract #{contract_id} and link #{link_name} already exists and must be deleted first"
+           )}
 
         _ ->
           {:reply, failure_ack("Invalid link definition put request")}
@@ -588,7 +591,9 @@ defmodule HostCore.ControlInterface.Server do
     })
   end
 
-  defp has_keys?(keys, request) do
+  defp has_keys?(keys, request) when is_map(request) do
     Enum.all?(keys, &Map.has_key?(request, &1))
   end
+
+  def has_keys(_keys, _request), do: false
 end
--- a/host_core/lib/host_core/control_interface/server.ex
+++ b/host_core/lib/host_core/control_interface/server.ex
@@ -143,22 +143,24 @@ defmodule HostCore.ControlInterface.Server do
   # Put a link definition
   # This will first store the link definition in memory, then publish it to the stream
   # then publish it directly to the relevant provider via the RPC channel
+  @put_linkdef_keys ["actor_id", "contract_id", "link_name", "provider_id", "values"]
   defp handle_request({"linkdefs", "put"}, body, _reply_to) do
     Tracer.with_span "Handle Linkdef Put (ctl)", kind: :server do
       with {:ok, ld} <- Jason.decode(body),
-           true <-
-             ["actor_id", "contract_id", "link_name", "provider_id"]
-             |> Enum.all?(&Map.has_key?(ld, &1)) do
-        HostCore.Linkdefs.Manager.put_link_definition(
-          ld["actor_id"],
-          ld["contract_id"],
-          ld["link_name"],
-          ld["provider_id"],
-          ld["values"] || %{}
-        )
-
+           true <- has_keys?(@put_linkdef_keys, ld),
+           :ok <-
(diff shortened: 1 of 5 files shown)


host_core/test/e2e/echo_test.exs
```diff
@@ -111,15 +113,6 @@ defmodule HostCore.E2E.EchoTest do
         @httpserver_key
       )
 
-    # For now, okay to put a link definition without proper claims
```
drowzy, line 114: I was unsure why this needed to be called again, so if I miss-understood the test scenario let me know :).

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -143,22 +143,24 @@ defmodule HostCore.ControlInterface.Server do
...
+               ld["actor_id"],
+               ld["contract_id"],
+               ld["link_name"],
+               ld["provider_id"],
+               ld["values"]
+             ) do
         {:reply, success_ack()}
       else
+        {:error, {:duplicate_key, {_, _, link_name}}} ->
+          {:reply, failure_ack("Duplicate link definition entry for '#{link_name}'")}
```
brooksmtownsend, line 162: I think this failure ack would be more informative as
```
Link definition for {actor_id} on contract {contract_id} and link {link_name} already exists and must be deleted first
```

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -586,4 +587,8 @@ defmodule HostCore.ControlInterface.Server do
       error: error
     })
   end
+
+  defp has_keys?(keys, request) do
+    Enum.all?(keys, &Map.has_key?(request, &1))
+  end
```
brooksmtownsend, line 593: Just for safety, since the `request` is user input here, I'd recommend adding a `when is_map(request)` to this function and setting up another function that handles the fallback case
```
defp has_keys?(keys, request), do: false
```

host_core/lib/host_core/control_interface/server.ex
brooksmtownsend: It's a little wordy but I'd rather include all the information than not enough

host_core/test/e2e/echo_test.exs
brooksmtownsend: I think this was just a leftover copy/paste, good catch 😄

host_core/test/e2e/echo_test.exs
```diff
@@ -5,7 +5,9 @@ defmodule HostCore.E2E.EchoTest do
   setup do
     {:ok, evt_watcher} =
       GenServer.start_link(HostCoreTest.EventWatcher, HostCore.Host.lattice_prefix())
-
+    on_exit(fn ->
+      :ets.delete_all_objects(:linkdef_table)
+    end)
```
brooksmtownsend, line 10: The EventWatcher will actually take care of this on initialization of each test, it's a fine safety measure but not necessary

host_core/lib/host_core/linkdefs/manager.ex
```diff
@@ -34,19 +38,20 @@ defmodule HostCore.Linkdefs.Manager do
...
+        provider_id: provider_key,
+        link_name: link_name,
+        contract_id: contract_id,
+        values: values,
+        deleted: false
+      }
 
-    publish_link_definition(ld)
+      publish_link_definition(ld)
+    end
```
brooksmtownsend, line 54: TIL this returns the value that doesn't match `:ok` as a default with no `else` required

brooksmtownsend (changes requested):
A couple of nitty requests, but overall this looks great!

host_core/lib/host_core/control_interface/server.ex
drowzy: Agree

brooksmtownsend (approved):
This worked well for me (actor ID overflows but we should address that in the actual modal, not a part of this PR)
image

brooksmtownsend (changes requested):
@drowzy Thank you for this contribution again, I'm only requesting changes since we've identified that this will significantly impact some edge cases around creating link definitions and I want to make sure we've thought those through before merging and releasing. No action required here on your part, we'll likely come back to this shortly and merge it once we can identify how this will change the runtime 😄

brooksmtownsend:
Since this has been open for a while and the linkdefs manager has changed significantly, going to close this for now.

PR #460, opened 2022-09-04
Return better error message when call alias is not found

### LINKED ISSUE #419: [BUG] Missing link definition error shows for a missing call alias
**Describe the bug**

When trying to actor-to-actor call a call alias that doesn't exist I get the following error:
```
15:51:14.038 [error] Invocation failure: Guest call failed for method HttpServer.HandleRequest: Host send error Invocation not authorized: missing link definition for petclinic/vets on default
```

This is a little misleading as the actual issue here is that there's no actor with the call alias `petclinic/vets` running. I think in this case we should be able to disambiguate an actor-to-actor call from a missing link definition

mattwilkinsonn:
Fixes #419.

Short circuits the piping in `host_call` to return a better error message when `identify_target` fails to find a target entirely.

I considered turning the entire pipe into the `with` syntax, but that looked too scary for my first `wasmcloud-otp`/Elixir change 😅. Let me know if there's a better way to do this.

Signed-off-by: Matt Wilkinson <matt@mattwilkinson.dev>

### DIFF
--- a/host_core/lib/host_core/web_assembly/imports.ex
+++ b/host_core/lib/host_core/web_assembly/imports.ex
@@ -230,7 +230,7 @@ defmodule HostCore.WebAssembly.Imports do
     if target == :unknown do
       {:error, :alias_not_found, token}
     else
-      {:ok, Map.put(token, :target, target)}
+      {:ok, %{token | target: target}}
     end
   end
 
--- a/host_core/lib/host_core/web_assembly/imports.ex
+++ b/host_core/lib/host_core/web_assembly/imports.ex
@@ -156,25 +156,36 @@ defmodule HostCore.WebAssembly.Imports do
       Tracer.set_attribute("payload_size", byte_size(payload))
 
       res =
-        %{
-          payload: payload,
-          binding: binding,
-          namespace: namespace,
-          operation: operation,
-          seed: seed,
-          claims: claims,
-          prefix: prefix,
-          state: state,
-          agent: agent,
-          source_actor: actor,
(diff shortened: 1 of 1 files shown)


host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -216,7 +227,11 @@ defmodule HostCore.WebAssembly.Imports do
           end
       end
 
-    %{token | target: target}
+    if target == :unknown do
+      {:error, :alias_not_found, token}
+    else
+      {:ok, Map.put(token, :target, target)}
```
brooksmtownsend, line 233: They are equivalent but just to keep with the style in this file I'd recommend doing
```elixir
{:ok, %{token | target: target}}
```

brooksmtownsend: This `put`s the key in the map and results in the new map

brooksmtownsend (commented):
This is a great start to refactoring one of the more complex areas of the host, I like the idea of continuing to improve this with `with` statements to make this more readable and error-message-friendly

brooksmtownsend (commented):
FWIW I think that this little block is better suited for a `with` statement than the piping like we were doing, the piping is nice but being able to individually segment steps is more beneficial to this piece of logic. 

I'd be happy to collaborate on converting this whole thing to a `with` statement, but this PR is fine fixing one issue at a time 😄

brooksmtownsend (approved):
Great first step in refactoring this statement

PR #464, opened 2022-09-09
add more validation to control interface server

connorsmith256:

### DIFF
--- a/host_core/lib/host_core/control_interface/server.ex
+++ b/host_core/lib/host_core/control_interface/server.ex
@@ -161,9 +161,7 @@ defmodule HostCore.ControlInterface.Server do
   defp handle_request({"linkdefs", "put"}, body, _reply_to) do
     Tracer.with_span "Handle Linkdef Put (ctl)", kind: :server do
       with {:ok, ld} <- Jason.decode(body),
-           true <-
-             ["actor_id", "contract_id", "link_name", "provider_id"]
-             |> Enum.all?(&Map.has_key?(ld, &1)) do
+           true <- has_values(ld, ["actor_id", "contract_id", "link_name", "provider_id"]) do
         HostCore.Linkdefs.Manager.put_link_definition(
           ld["actor_id"],
           ld["contract_id"],
@@ -187,9 +185,7 @@ defmodule HostCore.ControlInterface.Server do
   defp handle_request({"linkdefs", "del"}, body, _reply_to) do
     Tracer.with_span "Handle Linkdef Del (ctl)", kind: :server do
       with {:ok, ld} <- Jason.decode(body),
-           true <-
-             ["actor_id", "contract_id", "link_name"]
-             |> Enum.all?(&Map.has_key?(ld, &1)) do
+           true <- has_values(ld, ["actor_id", "contract_id", "link_name"]) do
         HostCore.Linkdefs.Manager.del_link_definition(
           ld["actor_id"],
           ld["contract_id"],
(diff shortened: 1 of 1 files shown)


host_core/lib/host_core/control_interface/server.ex
```diff
@@ -198,8 +194,7 @@ defmodule HostCore.ControlInterface.Server do
   # %{"actor_ref" => "bindle://example.com/echo/0.12.0", "host_id" => "Nxxxx", "count" => 4}
   defp handle_request({"cmd", _host_id, "la"}, body, _reply_to) do
     with {:ok, start_actor_command} <- Jason.decode(body),
-         true <-
-           Map.has_key?(start_actor_command, "actor_ref") do
+         actor_ref <- Map.get(start_actor_command, "actor_ref") do
```
brooksmtownsend, line 197: No matter the contents of the map, this will always succeed even if actor_ref will be nil. Just a note, you don't need it in the `with` statement, but it might be worth checking this with an `is_nil` or something if that's the goal

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -328,14 +317,14 @@ defmodule HostCore.ControlInterface.Server do
                 start_provider_from_bindle(
                   start_provider_command["provider_ref"],
                   start_provider_command["link_name"],
-                  Map.get(start_provider_command, "configuration", ""),
+                  Map.get(start_provider_command, "configuration") || "",
```
brooksmtownsend, line 320: These two are functionally equivalent, right? Just wondering the motivation for the change

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -586,4 +569,9 @@ defmodule HostCore.ControlInterface.Server do
       error: error
     })
   end
+
+  # returns true if all keys have non-nil values in the map
+  defp has_values(m, keys) when is_map(m) and is_list(keys) do
+    Enum.all?(keys, &Map.get(m, &1))
+  end
```
brooksmtownsend, line 576: It might be worth it to have a fallback case that just returns false, just in case we screw up the typing.

Either that, or, a dialyzer `@spec` so that we get warnings when we violate the typing, since this function is called from contexts where we are pretty sure of the data shape

host_core/lib/host_core/control_interface/server.ex
connorsmith256: Oops! Good catch.

host_core/lib/host_core/control_interface/server.ex
connorsmith256: I can't remember 🤔 I'll revert this and the change below.

host_core/lib/host_core/control_interface/server.ex
```diff
@@ -194,7 +194,8 @@ defmodule HostCore.ControlInterface.Server do
   # %{"actor_ref" => "bindle://example.com/echo/0.12.0", "host_id" => "Nxxxx", "count" => 4}
   defp handle_request({"cmd", _host_id, "la"}, body, _reply_to) do
     with {:ok, start_actor_command} <- Jason.decode(body),
-         actor_ref <- Map.get(start_actor_command, "actor_ref") do
+         actor_ref <- Map.get(start_actor_command, "actor_ref"),
+         true <- actor_ref != nil do
```
brooksmtownsend, line 198: There's actually a builtin function `is_nil` that you can use for this too 😄 https://hexdocs.pm/elixir/1.12/Kernel.html#is_nil/1

brooksmtownsend (approved):

connorsmith256:
Had merge conflicts :(

brooksmtownsend (approved):

PR #465, opened 2022-09-12
Adding WASI support

autodidaddict:
This PR adds preliminary WASI support, with the following limitations:
* emissions to stdout and stderr currently go into the void (subsequent work can poll them and forward to logger)
* WASI arguments (e.g. psuedo commandline) are empty
* WASI environment variables are empty
* WASI pre-vetted files and directories are empty

The net result of this PR is that WASI modules won't break when loaded, shouldn't have runtime failures, and can access functionality like the clock and random number generator.

Documentation will be forthcoming that outlines the caveats associated with using WASI-based actors.

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -131,7 +131,6 @@ defmodule HostCore.Actors.ActorModule do
       {:ok, stdout} = Wasmex.Pipe.create()
       {:ok, stderr} = Wasmex.Pipe.create()
 
-
       wasi = %{
         args: [],
         env: %{},
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -432,7 +432,7 @@ defmodule HostCore.Actors.ActorModule do
       end
 
     case Wasmex.start_link(opts)
-         |> prepare_module(agent, oci, false) do
+         |> prepare_module(agent, oci, true) do
       {:ok, agent} ->
         Agent.update(agent, fn state ->
           %State{state | ociref: oci}
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -432,7 +432,7 @@ defmodule HostCore.Actors.ActorModule do
       end
 
     case Wasmex.start_link(opts)
-         |> prepare_module(agent, oci) do
+         |> prepare_module(agent, oci, false) do
       {:ok, agent} ->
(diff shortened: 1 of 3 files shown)


autodidaddict:
bumped into an edge case that didn't show up during manual testing. Will get back to this as time permits.

autodidaddict:
This errors when the assumption was that it would work quietly: https://github.com/tessi/wasmex/blob/f4894f78b9854c7256f9c7265add5e0187c3a546/test/wasi_test.exs#L18

brooksmtownsend (changes requested):
I was able to start the echo actor fine, do we want to consider transforming one of our test fixture actors to be a WASI compiled actor so we don't accidentally regress here?

Adding another test for WASI support would be helpful too, so that we can verify this works automatically

brooksmtownsend (dismissed):
Big fan of having a test, and as we get further into WASI things we can figure out how to best test the WASI specific functionality.

Just to throw it out there, using an echo WASI actor like this I get about 3-5x faster benchmarks on fast requests 😄 
1 actor, 1 parallel request, as fast as possible
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        4.73 K      211.37 μs    ±60.83%      199.58 μs      410.88 μs
nats_echo_request          3.11 K      321.64 μs    ±63.17%      302.92 μs      731.80 μs
```
1 actor, 10 parallel requests
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        449.41        2.23 ms    ±24.13%        2.08 ms        4.21 ms
nats_echo_request          372.59        2.68 ms    ±57.94%        2.52 ms        4.57 ms
```
100 actors, 10 parallel requests
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        1.40 K        0.72 ms    ±58.84%        0.63 ms        1.90 ms
nats_echo_request          0.94 K        1.06 ms    ±54.90%        1.01 ms        1.77 ms
```
10 actors, 1 parallel request
```
Name                          ips        average  deviation         median         99th %
direct_echo_request        4.70 K      212.55 μs    ±33.58%      204.67 μs      357.99 μs
nats_echo_request          3.18 K      314.91 μs    ±31.44%      308.88 μs      456.90 μs
```

brooksmtownsend (dismissed):
Not a big deal but the echo wasi actor is 12.3MB :O Is that just a debug actor?

autodidaddict:
> Not a big deal but the echo wasi actor is 12.3MB :O Is that just a debug actor?

yes it was compiled in debug mode. I wanted the symbols in there in case it blew up on regression

brooksmtownsend (approved):

PR #468, opened 2022-09-15
Puts js_domain in host info for providers. Adds cluster key (issuer) to host probes

autodidaddict:
1. This adds the `js_domain` field to the host information data payload sent to providers during startup. This is only useful if the provider (or underlying RPC library) needs to access things like jetstream for RPC "chunking", etc.
2. Adds the `issuer` field to the host ping response
3. Adds the `issuer` field to the host inventory response

This shouldn't break the control interface clients written in Rust because serde will ignore the unknown field(s). However, @jordan-rash this might break the corresponding Go client (I'm not sure how its JSON deserialization handles unexpected fields)

### DIFF
--- a/host_core/lib/host_core.ex
+++ b/host_core/lib/host_core.ex
@@ -196,8 +196,7 @@ defmodule HostCore do
 
             %{
               config
-              | cluster_seed: config.cluster_seed,
-                cluster_key: config.cluster_key,
+              | cluster_key: pk,
                 cluster_issuers: issuers,
                 cluster_adhoc: false
             }
--- a/host_core/lib/host_core/host.ex
+++ b/host_core/lib/host_core/host.ex
@@ -46,6 +46,7 @@ defmodule HostCore.Host do
     friendly_name = HostCore.Namegen.generate()
 
     Logger.info("Host #{opts.host_key} (#{friendly_name}) started.")
+    Logger.info("Host issuer public key: #{opts.cluster_key}")
     Logger.info("Valid cluster signers: #{opts.cluster_issuers}")
 
     if opts.cluster_adhoc do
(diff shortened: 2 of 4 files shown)


jordan-rash:
Thanks for the heads up @autodidaddict

host_core/lib/host_core.ex
```diff
@@ -196,8 +196,7 @@ defmodule HostCore do
 
             %{
               config
-              | cluster_seed: config.cluster_seed,
```
brooksmtownsend, line 199: Does this mean we won't be able to refer to a host_config file to use generated cluster seeds for subsequent hosts?

host_core/lib/host_core/host.ex
```diff
@@ -46,6 +46,7 @@ defmodule HostCore.Host do
     friendly_name = HostCore.Namegen.generate()
 
     Logger.info("Host #{opts.host_key} (#{friendly_name}) started.")
+    Logger.info("Host issuer public key: #{opts.cluster_key}")
```
brooksmtownsend, line 49: I was going to call for a clarification in language here, but if you search `issuer` on our docsite it comes to https://wasmcloud.dev/app-dev/secure/clusterkeys/ so this is good

brooksmtownsend (changes requested):
The provider changes look good, but I don't want to lose the ability to refer to a host_config.json to launch a host with a defined cluster seed

host_core/lib/host_core.ex
autodidaddict: no, this was a bug that basically used to say `config.cluster_seed: config.cluster_seed` which is dum. Now it's using the right value :)

brooksmtownsend (dismissed):

autodidaddict:
> The provider changes look good, but I don't want to lose the ability to refer to a host_config.json to launch a host with a defined cluster seed

This won't happen. You can still supply a seed via environment variable and have one in a host_config.json file. In my tests I re-ran the same host multiple times and it had the same seed (and issuer) each time. Deleted the host_config.json and it got a new seed. supplied one via env vars and it honored that one.

brooksmtownsend (approved):

PR #470, opened 2022-09-16
style(web): add values column in link def table

rovast:
Signed-off-by: rovast <rovast@163.com>

Add a value column in the link definition table to check the definition information on the dashboard.

image

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host_web/live/components/link_row_component.ex
+++ b/wasmcloud_host/lib/wasmcloud_host_web/live/components/link_row_component.ex
@@ -19,6 +19,7 @@ defmodule LinkRowComponent do
     <tr>
       <td><%= @link_name %></td>
       <td><%= @contract_id %></td>
+      <td><%= str_values(@values) %></td>
       <td><button class="btn btn-primary btn-sm id-monospace" type="button"
           onClick="navigator.clipboard.writeText('<%= @actor_id %>')" data-toggle="popover" data-trigger="focus" title=""
           data-content="Copied!">
@@ -49,5 +50,11 @@ defmodule LinkRowComponent do
       </td>
     </tr>
     """
+  end
+
+  defp str_values(value) do
+    value
+    |> Enum.map(fn {a, b} -> "#{a}=#{b}" end)
+    |> Enum.join(",")
   end
 end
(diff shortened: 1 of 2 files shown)


brooksmtownsend (changes requested):
@rovast I really like the idea behind this PR, however I have practical concerns for displaying link values alongside the dashboard. For example, if you were to link an actor to a blobstore with AWS credentials, this is what it may look like
image

I think, practically, it will be really difficult to design a UI that can handle longer link values, especially the base64 or JSON encoded ones.

One thing I would support, though, is a button that allows you to copy link values to the clipboard. This way we don't have to design the UI for it and you can always retrieve link values if desired. 

Thoughts @autodidaddict / @connorsmith256 ?

connorsmith256:
Agreed with your comments @brooksmtownsend

I'm interested in having a broader conversation about sensitive values in link defs, but for the time being a copy button might suffice

rovast:
Yes! That's really a good idea, and if the values map is empty, no copy button is shown.

image

wasmcloud_host/lib/wasmcloud_host_web/live/components/link_row_component.ex
```diff
@@ -19,6 +19,15 @@ defmodule LinkRowComponent do
     <tr>
       <td><%= @link_name %></td>
       <td><%= @contract_id %></td>
+      <td><%= if map_size(@values) > 0 do %><button class="btn btn-primary btn-sm id-monospace" type="button"
```
brooksmtownsend, line 22: Just a nit, can this button be centered to keep it in the middle of the column?

brooksmtownsend (changes requested):

brooksmtownsend (changes requested):
Now that I'm looking at the screenshot, I actually think the values copy button should be alongside the actions since it's something you can do with the linkdef. Thoughts? It would be best if it also had that popover that showed the tooltip for "Copy Link Definition Values" as well

rovast:
1. move the copy value button to the action column

image

2. add tooltip for both delete and copy button

image

brooksmtownsend (commented):
This looks great! I'll test it first thing when I get on tomorrow AM.

@connorsmith256 @autodidaddict do you have any other thoughts here?

brooksmtownsend (approved):
Looks great!

PR #471, opened 2022-09-16
(Helm) Allow insecure registries and latest tag, for local/homeserver development

handicraftsman:
Just as the PR name states, this is only for development purposes.

My scenario for insecure registries on helm:

I run a home server with microk8s, on a desktop computer, and an insecure registry there. \
It is then used to push images from my laptop, and pull into the wasmcloud deployment. \
Without these changes I cannot deploy actors/providers in any way other than uploading manually, with a GUI.


brooksmtownsend (approved):
This looks great 😄 Thanks @handicraftsman for this contribution!

PR #481, opened 2022-10-12
Update otel docker-compose to support seach traces

rovast:
The new docker-compose example supports search traces. So that we do not need to inspect the otel docker log or wasmhost log to get traceID.

image

### DIFF
--- a/otel/compose.yaml
+++ b/otel/compose.yaml
@@ -1,10 +1,11 @@
 services:
   grafana:
-    image: grafana/grafana:8.1.2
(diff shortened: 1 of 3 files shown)


thomastaylor312:
Hey @rovast! This is really great! Code changes look straightforward, but I think we just want to test it to double check things. We're all a bit slammed with work right now so it might be a little bit longer than normal to get you a review, but we'll do it as soon as possible.

brooksmtownsend (approved):
image

This worked great!

PR #483, opened 2022-10-13
replace nil with empty binary on 'msg' field of invocation responses

connorsmith256:
When wasmbus-rpc attempts to deserialize the MessagePack envelope, nils fail to deserialize to an empty Vec. This was causing a bunch of illegible error messages any time the invocation response didn't include a message


stevelr (approved):

brooksmtownsend (approved):
Yay

PR #486, opened 2022-10-14
fix: try pinning versions

ricochet:


ricochet:
We moved on to Brooks's branch

PR #489, opened 2022-10-19
Attempt to create consumer regardless of stream response

brooksmtownsend:
This PR removes the "fail early" aspect of the jetstream client. In testing we've seen a few different error messages come from the NATS stream create message, and because we were short-circuiting if we received an unexpected response (even if it was OK to proceed) we broke some new emerging scenarios with stream creates.

After this PR, even if the stream fails to create we'll attempt to create the consumer anyways. Some valid reasons for this include:
1. Stream already exists with that name
    1. We proceed safely and create the ephemeral consumer 
3. Stream already exists with that name with a different configuration
    1. We proceed safely and attempt to create the ephemeral consumer (which should not fail, unless there's a max number of consumers)
4. The NATS user is allowed to subscribe to streams but not create them
    1. We proceed safely and create the ephemeral consumer

If the stream create fails, we `warn` and proceed regardless to create the consumer, which will error if anything goes wrong. This won't prevent the host from launching, but we may end up closing the host in this case

More scenarios probably fall under this umbrella, and in the interest of not updating the host for each emerging scenario we can just attempt to proceed with creating the ephemeral consumer. Feasibly, if this is a legitimate error scenario, we'll fail for the same reason in both places and _then_ modification of NATS config or host handlers could be required.

### DIFF
--- a/host_core/lib/host_core/jetstream/client.ex
+++ b/host_core/lib/host_core/jetstream/client.ex
@@ -51,29 +51,20 @@ defmodule HostCore.Jetstream.Client do
       }
       |> Jason.encode!()
 
-    cont =
-      case HostCore.Nats.safe_req(:control_nats, create_topic, payload_json) do
-        {:ok, %{body: body}} ->
-          handle_stream_create_response(body |> Jason.decode!())
+    case HostCore.Nats.safe_req(:control_nats, create_topic, payload_json) do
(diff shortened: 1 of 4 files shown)


autodidaddict (approved):
LGTM, would like to see how this reacts in a couple of these scenarios

autodidaddict (approved):

PR #490, opened 2022-10-19
Removed host_core release from actions

brooksmtownsend:
As of the `0.58.0` release, we're seeing the host_core Dockerfile fail to build. This is unfortunate, and with an opaque error message difficult to debug. This PR removes this build from our release pipeline, for now.

This is acceptable only because the host_core docker image and tarball don't provide the wasmCloud experience that we want, at this moment. Capability providers that are destined for this image (a musl libc image) need to be compiled _specifically_ with this in mind which is a lot to ask of both our users and ourselves. Until we can transition to architecture-agnostic capability providers, the complexity of compiling our Wasmex package manually and maintaining a Docker image isn't reasonable for the benefit.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -171,82 +171,82 @@ jobs:
           name: ${{ matrix.release-tarball }}
           path: ${{env.working-directory}}/${{ matrix.release-tarball }}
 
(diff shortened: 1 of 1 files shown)


connorsmith256 (approved):
😢

PR #492, opened 2022-10-28
Fixing orphan processes during actor shutdown

autodidaddict:
This change resolves a subtle memory leak. Every time we create an actor we create two child processes - an agent for state, and a `wasmex` genserver for the webassembly module itself. Apparently when using `:normal` as a termination reason, children created with `start_link` are orphaned rather than terminated. This means that the memory consumed by those orphaned children isn't reclaimed.

This changes the shutdown reason to `:shutdown`, which will explicitly cause the GenServer to terminate children and therefore be able to reclaim the resources used by them.

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -221,7 +221,11 @@ defmodule HostCore.Actors.ActorModule do
 
     publish_actor_stopped(public_key, instance_id)
 
-    {:stop, :normal, :ok, agent}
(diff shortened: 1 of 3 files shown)


brooksmtownsend (commented):
@autodidaddict would you mind putting a little more detail into the description as to what this is solving? Additionally, are there any other places where we might employ this? I know that providers have the same `{:stop, normal, ...` result

brooksmtownsend (approved):

connorsmith256 (approved):

PR #493, opened 2022-10-28
Fix/cache actor modules

connorsmith256:
This adds a `:module_cache` ets table used to cache the compiled bytes for an actor. I had a suspicion re-compiling on every actor instantiation was contributing to the host consuming more memory than it should.

I tested this change with the following script:

```bash
host_id=$(wash ctl get hosts -o json | jq -r '.hosts | .[0].id')
actor_id=MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5
actor_ref=wasmcloud.azurecr.io/echo:0.3.5
N=10

while true; do
    for i in $(seq 1 $N); do
        wash ctl scale actor $host_id $actor_id $actor_ref --count $i
    done
done
```

On main, if you leave this running, the host's memory will quickly grow, and grow, and grow, and.... With this change, the host uses less memory per instance, and it also reclaims nearly(?) all memory when an instance is terminated

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -411,11 +411,13 @@ defmodule HostCore.Actors.ActorModule do
       wasmbus: Imports.wasmbus_imports(agent)
     }
 
-    module = case :ets.lookup(:module_cache, claims.public_key) do
-      [{_, cached_mod}] -> cached_mod
+    key = :crypto.hash(:sha256, bytes) |> Base.encode16() # we consider a hash of bytes as a unique key
+    module = case :ets.lookup(:module_cache, key) do
+      [{_, cached_mod}] ->
+        cached_mod
       [] ->
         {:ok, mod} = Wasmex.Module.compile(bytes)
-        :ets.insert(:module_cache, {claims.public_key, mod})
+        :ets.insert(:module_cache, {key, mod})
         mod
     end
 
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -411,15 +411,19 @@ defmodule HostCore.Actors.ActorModule do
       wasmbus: Imports.wasmbus_imports(agent)
     }
 
-    key = :crypto.hash(:sha256, bytes) |> Base.encode16() # we consider a hash of bytes as a unique key
-    module = case :ets.lookup(:module_cache, key) do
-      [{_, cached_mod}] ->
-        cached_mod
-      [] ->
-        {:ok, mod} = Wasmex.Module.compile(bytes)
-        :ets.insert(:module_cache, {key, mod})
-        mod
-    end
+    # we consider a hash of bytes as a unique key
+    key = :crypto.hash(:sha256, bytes) |> Base.encode16()
+
+    module =
+      case :ets.lookup(:module_cache, key) do
+        [{_, cached_mod}] ->
+          cached_mod
+
+        [] ->
+          {:ok, mod} = Wasmex.Module.compile(bytes)
+          :ets.insert(:module_cache, {key, mod})
+          mod
+      end
 
     # TODO - in the future, poll these so we can forward the err/out pipes
     # to our logger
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -407,7 +411,13 @@ defmodule HostCore.Actors.ActorModule do
...
     }
 
-    {:ok, module} = Wasmex.Module.compile(bytes)
+    module = case :ets.lookup(:module_cache, claims.public_key) do
+      [{_, cached_mod}] -> cached_mod
+      [] ->
+        {:ok, mod} = Wasmex.Module.compile(bytes)
+        :ets.insert(:module_cache, {claims.public_key, mod})
+        mod
+    end
```
autodidaddict, line 420: This is amazing

autodidaddict (approved):

connorsmith256:
@brooksmtownsend pointed out offline that we shouldn't use just the actor's PK as the cache key, since that will remain the same during local dev iteration. Instead we should use the actor hash in the claims

autodidaddict:
The edge case we want to avoid is maintaining many many many caches for every iteration locally. That cache will grow when instead we want to be replacing the compiled module that corresponds to the public key

autodidaddict:
The edge case we want to avoid is maintaining many many many caches for every iteration locally. That cache will grow when instead we want to be replacing the compiled module that corresponds to the public key

A live update should also clear the cached module.

stevelr:
Some ideas on cache key:

- if the module was loaded from a file path:
    path + last_modified_time
- if from oci url:
   url + insecure-flag + user + password 
   (name+pw ensures that you need to have the oci creds to access a cached module from a private registry) 
   
For both cases, also add:
- host's build version
- host arch, sys_arch, and os_type

Don't include lattice_id, if you want to share cached artifacts between lattices, or add lattice_id if artifacts should never be shared.

Rust code sample:
```
pub fn cache_id(&self, lattice: &Lattice, origin: &ModuleOrigin) -> Option<String> {
        let mut hasher = sha2::Sha256::new();
        match &origin {
            ModuleOrigin::Path(path) => {
                hasher.update(format!("{}", path.display()));
                hasher.update(file_age(path).to_le_bytes());
            }
            ModuleOrigin::Oci(oci_url) => {
                hasher.update(oci_url);
                if let Some(oci_auth) = lattice.oci_auth_config(oci_url) {
                    hasher.update(oci_auth.domain);
                    hasher.update(oci_auth.insecure.to_string());
                    hasher.update(oci_auth.user.unwrap_or_default());
                    hasher.update(oci_auth.password.unwrap_or_default());
                } else {
                    hasher.update("no.auth");
                }
            }
            _ => return None,
        }
        // host version
        hasher.update(env!("CARGO_PKG_VERSION").as_bytes());
        // compilation results depend on architecture ...
        hasher.update(crate::arch::sys_os_family().as_bytes());
        hasher.update(crate::arch::sys_arch().as_bytes());
        hasher.update(crate::arch::sys_os_type().as_bytes());
        Some(hex::encode(hasher.finalize()))
    }
```

autodidaddict:
This seems super complicated. We should be able to come up with a simpler key or just invalidate the cache when appropriate.

stevelr:
>  just invalidate the cache when appropriate.
That's the tricky part.
The algorithm above based on what wasmtime does, with the addition of oci creds.

stevelr:
I can explain a justification for each line above. I think it's close to as simple as it can be while still being accurate.

autodidaddict:
Bytes can come from arbitrary locations, which may not include a file path or even an oci reference. 

The only thing this cache is attempting to accomplish is to store an in memory copy of a given module. 

If we're not using the cache to avoid downloads (a different concern), then Brooks idea is probably the simplest just using the hash of the bytes.

stevelr:
that's fair.
The purpose of this id is to avoid downloads + avoid redundant AOT compilation, for faster startup time.

brooksmtownsend (approved):
@autodidaddict if you merge #492 then this branch can rebase on main

brooksmtownsend (approved):

PR #494, opened 2022-11-07
Add host ID to provider shutdown message

autodidaddict:
This code adds the host ID of the host that handled the terminate provider message to the shutdown message sent to the provider. This (in another PR) will allow the provider to decide if it was the true target of the shutdown. If it wasn't the target, it will stay running.

Testing Method:
* Updated the HTTP server interface to use new code
* Updated the HTTP server capability provider to use the new interface
* Updated the wasmCloud OTP host (this PR) to include the target host ID on the shutdown message
* Start 2 http server providers , 1 per host
* Issue a stop command for one host
* Observe 1 provider still running

Screenshot 2022-11-07 at 9 15 21 AM

### DIFF
--- a/host_core/mix.exs
+++ b/host_core/mix.exs
@@ -1,7 +1,7 @@
 defmodule HostCore.MixProject do
   use Mix.Project
 
-  @app_vsn "0.58.4"
+  @app_vsn "0.58.3"
 
   def project do
     [
--- a/host_core/mix.exs
+++ b/host_core/mix.exs
@@ -1,7 +1,7 @@
 defmodule HostCore.MixProject do
   use Mix.Project
 
-  @app_vsn "0.58.3"
+  @app_vsn "0.58.4"
 
   def project do
     [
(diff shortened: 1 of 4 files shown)


host_core/mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule HostCore.MixProject do
   use Mix.Project
 
-  @app_vsn "0.58.3"
+  @app_vsn "0.58.4"
```
brooksmtownsend, line 4: Make sure to bump this version in `wasmcloud_host` and in the Chart as well

brooksmtownsend (changes requested):
Version change request.

~Checking for understanding, so this does require our capability providers to upgrade as well, right? This change may be better handled in `wasmbus-rpc` so that capability providers can get this by default even if we don't write them~

edit: I am teh dumb, found linked PR

host_core/mix.exs
brooksmtownsend: @autodidaddict I think this can just be `0.58.3` since we never released it, looks like it was bumped separately before

brooksmtownsend (approved):

connorsmith256 (approved):

PR #495, opened 2022-11-08
Initial support for lattice-local artifact storage

autodidaddict:
This adds support for the "launch actor" control interface command to contain actor references with the scheme/prefix of `lattice://`. When the host receives such an instruction, it will pull the bytes from a NATS object store named `ARTIFACT_{lattice}` and the object ID is the public key of the actor.

For example, if the actor module has a public key of `MAXGZOTHQGIISDCJQS577NY4NAFBOUI5MNFL43IMPEWX5V3YOGNYWO6D`, then you could send a request to launch this actor with the reference `lattice://MAXGZOTHQGIISDCJQS577NY4NAFBOUI5MNFL43IMPEWX5V3YOGNYWO6D`. The object must already exist in the `ARTIFACT_{lattice}` object store with the key `MAXGZOTHQGIISDCJQS577NY4NAFBOUI5MNFL43IMPEWX5V3YOGNYWO6D`.

_The host will not automatically create this object store_ because the host never writes to this store, it only ever reads. Therefore, tooling that writes to this store will be responsible for creation. (e.g. `wash`).

⚠️ **NOTE** - we are awaiting a change/fix to the NATS Rust client that deals with object name encoding. Until this release is adopted, objects uploaded using the `golang` client (which includes the `nats` CLI!) will not be visible to the host. Additionally, we've discovered that this same issue prevents `golang` capability providers from exchanging chunked invocations with the host.

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -146,6 +146,25 @@ defmodule HostCore.Actors.ActorSupervisor do
     end
   end
 
+  def start_actor_from_localstore(pk, count \\ 1, annotations \\ %{}) do
+    Tracer.with_span "Starting Actor from lattice object store", kind: :server do
+      case HostCore.WasmCloud.Native.get_actor_localobject(pk) do
+        {:error, error} ->
+          Tracer.add_event("Fetch from lattice object store failed", reason: "#{inspect(err)}")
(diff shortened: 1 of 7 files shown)


autodidaddict:
Looks like I had yet another git issue. stand by.

autodidaddict:
After some hands on time with this experiment, we've decided not to implement this as is

PR #496, opened 2022-11-09
Emit RPC events (invocation pass/fail) on a different topic for traffic segregation

autodidaddict:


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -690,7 +690,7 @@ defmodule HostCore.Actors.ActorModule do
       }
       |> CloudEvent.new(evt_type)
 
-    topic = "wasmbus.evt.#{prefix}"
```
connorsmith256, line 693: Should we make `wasmbus.evt` and `wasmbus.rpcevt` constant prefixes somewhere?

connorsmith256 (dismissed):

host_core/lib/host_core/actors/actor_module.ex
autodidaddict: eventually, yes... but I've got some thoughts for how to centralize publishing of events during the "big refactor" that we want to do.

autodidaddict: swapped to use module constants

connorsmith256 (approved):

brooksmtownsend:
Note that the decision implemented in this PR is documented in https://github.com/wasmCloud/adr/pull/13

PR #498, opened 2022-11-17
Add start provider configuration

rovast:
Some providers may need configuration to start, so we can input the configuration in UI, and pass the configuration to host_core.

image

### DIFF
--- a/host_core/lib/host_core/providers/provider_supervisor.ex
+++ b/host_core/lib/host_core/providers/provider_supervisor.ex
@@ -171,9 +171,16 @@ defmodule HostCore.Providers.ProviderSupervisor do
           host_id :: String.t(),
           path :: String.t(),
           link_name :: String.t(),
-          annotations :: map()
+          annotations :: map(),
+          provider_configuration :: String.t()
         ) :: DynamicSupervisor.on_start_child()
-  def start_provider_from_file(host_id, path, link_name, annotations \\ %{}) do
+  def start_provider_from_file(
+        host_id,
+        path,
+        link_name,
+        annotations \\ %{},
+        provider_configuration \\ ""
+      ) do
     Tracer.with_span "Start Provider from File" do
       case Native.par_from_path(path, link_name) do
         {:ok, par} ->
@@ -189,7 +196,7 @@ defmodule HostCore.Providers.ProviderSupervisor do
             link_name,
             par.contract_id,
             "",
-            "",
+            provider_configuration,
             annotations
           )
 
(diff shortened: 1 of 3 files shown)


autodidaddict:
This looks great!

connorsmith256 (dismissed):
Great addition! Should we support providing config to providers started from a file as well?

connorsmith256 (dismissed):

brooksmtownsend:
@rovast looks like this might just be missing something that's failing in the tests, probably the change in the number of arguments that go into `start_provider_from_file`

rovast:
> @rovast looks like this might just be missing something that's failing in the tests, probably the change in the number of arguments that go into `start_provider_from_file`

Help wanted. ✋ 

When I was developing this function, I found the same question the argument number does not match, and it confused me for a long while(That's why in the first version PR, [I ignore the support for the file](https://github.com/wasmCloud/wasmcloud-otp/pull/498#pullrequestreview-1184653513)). 

But after I changed the  `start_provider_from_file ` function, and reboot the host, the function works, and the issue is fixed. 

I'm not very familiar with Elixir, could you please edit the code directly? Thanks a lot.

brooksmtownsend:
@rovast sure thing, someone on the maintainer side should be able to tweak the arguments here in a bit and push to this PR. I appreciate you getting this feature so far and we're happy to just double check the Elixir bits 😄

connorsmith256:
@rovast @brooksmtownsend I think this was a case of flaky tests 😞 I noticed different tests failed on the 1st and 2nd run, and I just triggered a 3rd try and everything passed

@brooksmtownsend have we see flaky e2e tests before? The common failure between the two was trying to contact http://localhost:8081

brooksmtownsend:
> @brooksmtownsend have we see flaky e2e tests before? The common failure between the two was trying to contact http://localhost:8081/

Occasionally, yes, but I'm not sure exactly what happened in the previous run. I think we can give this a go locally and a few test runs in CI to make sure it's all good

brooksmtownsend:
@rovast I think the only thing we need here is a rebase, everything looks good!

rovast:
@brooksmtownsend ok, rebase is done, and all checks have passed. 😃

stevelr:
I like this! I have a few questions about implementation details:

Is the data type of provider_configuration a map or a string?

Should it be included when the host reports inventory? (What if the config includes secrets?)

Will it be added to HostData to send to providers on startup? I don't think I noticed a pr to the interface repo for the change to model-core.

connorsmith256:
@stevelr this change isn't adding a new kind of configuration, it's allowing users to supply the existing configuration format when starting providers from a file in the washboard

connorsmith256 (approved):
Sorry for the delay @rovast , and thank you for the contribution!

PR #499, opened 2022-11-18
fix(bindle): Makes bindle verification a little less strict

thomastaylor312:
We were trying to validate the creator signature as well by default. To keep it simple and transparent to users for now, I made it so we only validate the host signature. In the future we probably want to support a configurable value for the validation strategy


stevelr:
This change will be useful for https://github.com/Finfalter/wasmCloudArtefacts too.

autodidaddict (approved):

PR #504, opened 2022-11-25
Host_config.json double write fix and customization

### LINKED ISSUE #480: [BUG] host_config.json writes out to the current directory, not where the release is installed
**Describe the bug**

Before mix releases, the `host_config.json` file would write to the directory where the release is installed. Now, it's written wherever your `$(PWD)` reports when you run `./bin/wasmcloud_host start` or `wash up`. This results in JSON configuration files strewn about in the filesystem. This is especially frustrating with git since these files can alter host behavior and shouldn't be committed alongside other changes.

**To Reproduce**

1. Start host
2. See host_config.json

**Expected behavior**

I expect the `host_config.json` file to write out where the host release is unpacked, or with a configurable environment variable (perhaps an optional variable to avoid writing the file at all would be useful too)

**Environment (please complete the following information)**
 - OS: [e.g. Linux, MacOS, Windows] Linux/Macos, not tested on Windows
 - Arch: [e.g. x86_64, aarch64] aarch64
 - wasmCloud Version: [e.g. 0.51.0] 0.57.0+

giusdp:
Closes #480 

This PR removes the double write_json for the host_config.json file and adds a `HOST_CONFIG` env variable, so that it is possible to customize the configuration file to use. 

It adds `:host_config` bindings in config_plan.ex and a nil check in `post_process_config` in case the var was not given, in which case it defaults to `~/.wash/host_config.json`.

### DIFF
--- a/host_core/lib/application.ex
+++ b/host_core/lib/application.ex
@@ -23,7 +23,8 @@ defmodule HostCore.Application do
     :host_seed,
     :enable_structured_logging,
     :structured_log_level,
-    :host_key
+    :host_key,
+    :host_config
   ]
 
   def start(_type, _args) do
@@ -122,12 +123,30 @@ defmodule HostCore.Application do
         end
       end
 
+    host_config =
+      if is_nil(config.host_config) do
+        case System.user_home() do
+          nil ->
+            Logger.warn(
+              "Could not determine user's home directory. Using current directory instead."
+            )
+
+            @host_config_file
+
+          h ->
+            Path.join([h, "/.wash/", @host_config_file])
+        end
+      else
+        config.host_config
+      end
+
     config =
       config
       |> Map.put(:cluster_adhoc, false)
       |> Map.put(:cluster_key, "")
       |> Map.put(:host_key, host_key)
       |> Map.put(:host_seed, host_seed)
+      |> Map.put(:host_config, host_config)
 
     config = ensure_booleans(config)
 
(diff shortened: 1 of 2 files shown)


thomastaylor312 (approved):
Sorry for the delay @giusdp! This is looking good. Going to have @autodidaddict or @brooksmtownsend double check my review on this

Also, this will need a rebase to pick up the latest changes

host_core/lib/host_core.ex
```diff
@@ -129,12 +129,30 @@ defmodule HostCore do
         end
       end
 
+    host_config =
+      if config.host_config == nil do
```
brooksmtownsend, line 133: nit: it's a little more idiomatic to use `is_nil/1` here

brooksmtownsend (commented):
Just a small nit, and yeah @giusdp if you can do a quick rebase then this is good to go!

host_core/lib/host_core.ex
giusdp: Thanks! I used `is_nil/2` now

thomastaylor312 (approved):

giusdp:
I've done the rebase. The host_core.ex file changed name to application.ex so I've re-added my changes. I shouldn't have messed up anything (I hope), so let me know!

brooksmtownsend (approved):

brooksmtownsend:
Thanks @giusdp ! Looks great 😄

PR #511, opened 2022-12-05
Credo warnings for host_core/actors/ files

giusdp:
I'm starting to address some of the credo warnings. I focused the changes on the host_core/lib/host_core/actors files, using the command `mix credo --strict "lib/host_core/actors/*.ex"` from the host_core folder. 

I left out 3 warnings: 2 are about the #TODOs found in the code. I'd suggest these could be turned into issues here on github so the comments can be removed, fixing the credo warnings. The last warning is about the complexity of the `scale_actor` function, but I was unsure about changing the code too much.


brooksmtownsend (approved):
This all looks good to me, and a lot more readable for some of the places where we mixed `|>` and normal function calls! Thank you @giusdp

PR #512, opened 2022-12-06
Credo warnings for host_core/test files 

giusdp:
Another PR for more credo warnings. I focused the changes on the host_core/test files, using the command mix credo --strict "test/**/*ex" from the host_core folder.

I left out 1 TODO warning to not remove that comment. This PR removes a good chunk of the alias/nested modules readability warnings.

### DIFF
--- a/host_core/test/e2e/controlinterface_test.exs
+++ b/host_core/test/e2e/controlinterface_test.exs
@@ -1,6 +1,9 @@
 defmodule HostCore.E2E.ControlInterfaceTest do
   use ExUnit.Case, async: false
 
+  alias HostCore.Actors.ActorSupervisor
+  alias HostCore.Linkdefs.Manager
+
   import HostCoreTest.Common, only: [cleanup: 2, standard_setup: 1]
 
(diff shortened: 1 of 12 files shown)


brooksmtownsend (approved):
🚀 Love these improvements, looks like a few tests are failing in CI. @giusdp can you reproduce these failures locally or do all tests pass for you?

giusdp:
Locally the tests all pass. I'm not sure what's going on since I didn't touch any business logic. Maybe I could try smaller commits to check if there is a particular change that is making the CI fail.

autodidaddict:
> Locally the tests all pass. I'm not sure what's going on since I didn't touch any business logic. Maybe I could try smaller commits to check if there is a particular change that is making the CI fail.

I think what's happening is that some tests are more vulnerable to latency changes than others, and so sometimes those get caught on github.

I've got some changes on my own branch where I added `wait_for_actor_start` calls to the tests inside actor_test.exs this seems to be fine, though the only way to know if it'll be fine in github is to check those in.

PR #515, opened 2022-12-10
add valid return to publish_heartbeat cast

connorsmith256:
I noticed while running the host off `main` that it dies when responding to a control interface ping. I tracked it down to a bad return from handling a `:publish_heartbeat` cast


brooksmtownsend (approved):
Great catch, I wonder if this is causing some of the crashes we see with connections (the host trying to publish the heartbeat when it starts up with no NATS connection)

autodidaddict (approved):

PR #518, opened 2022-12-15
fix double host stop event 

### LINKED ISSUE #469: [BUG] A stopped host publishes two host_stopped events
**Describe the bug**

When you stop a wasmCloud host, it publishes two `host_stopped` events in the process of shutting down. With the [lattice-observer](https://github.com/wasmCloud/lattice-observer), this can end up double-handling the same event which is unnecessary

**To Reproduce**

1. Start a wasmcloud host
2. Use the NATS CLI to subscribe to `wasmbus.evt.*`
3. Stop it with `./bin/wasmcloud_host stop`
4. Observe two host_stopped events

**Expected behavior**

I'd expect a single `host_stopped` event to be published

giusdp:
This PR fixes #469 

The virtual_host.ex `handle_cast(:do_stop, _timeout_ms}, state)` function invokes `do_purge` and `publish_host_stopped`, then it returns the tuple `{:stop, :shutdown, state}`  triggering the `terminate\2` function.

In `terminate` those two functions are invoked again resulting in a double `host_stopped` event and a double purging of the processes (visible in the logs).

It should be enough not calling those functions immediately in the handle cast but letting `terminate` handle it.

### DIFF
--- a/host_core/lib/host_core/vhost/virtual_host.ex
+++ b/host_core/lib/host_core/vhost/virtual_host.ex
@@ -419,9 +419,6 @@ defmodule HostCore.Vhost.VirtualHost do
   def handle_cast({:do_stop, _timeout_ms}, state) do
     Logger.debug("Host stop requested manually")
 
-    do_purge(state)
-    publish_host_stopped(state)
-
     if HostCore.Application.host_count() == 1 do
       # Give a little bit of time for the event to get sent before shutting down
       :timer.sleep(300)


host_core/lib/host_core/vhost/virtual_host.ex
```diff
@@ -427,9 +427,6 @@ defmodule HostCore.Vhost.VirtualHost do
   def handle_cast({:do_stop, _timeout_ms}, state) do
     Logger.debug("Host stop requested manually")
 
-    do_purge(state)
-    publish_host_stopped(state)
-
```
brooksmtownsend, line 432: We may want to remove the `publish_host_stopped` call, but the `do_purge` is important since it instructs all providers to shutdown with some time to handle cleaning up state and things

brooksmtownsend (changes requested):

host_core/lib/host_core/vhost/virtual_host.ex
brooksmtownsend: Ah I see what you mean, that function ends up triggering the process termination. I think I misunderstood the change, and you are likely correct that both of those should be gone. Have you been able to test this either way with something like `nats sub wasmbus.evt.*` to listen for host_stopped events?

giusdp: I've re-added the `do_purge` removal. I checked with `iex -S mix` and `wash ctl stop` listening for the events and the host_stopped event appears only once. In the iex logs the purging works normally too.

brooksmtownsend (approved):

brooksmtownsend:
@giusdp sorry to bother you, I think the only thing we need here is a rebase and we're good to merge!

brooksmtownsend (approved):

brooksmtownsend:
Thanks for your patience @giusdp !

PR #520, opened 2022-12-16
Linkdefs put with a predictable hash ID rather than a random UUID

autodidaddict:


host_core/lib/host_core/linkdefs/manager.ex
```diff
@@ -168,6 +168,15 @@ defmodule HostCore.Linkdefs.Manager do
     end
   end
 
+  defp linkdef_hash(actor_id, contract_id, link_name) do
+    sha = :crypto.hash_init(:sha256)
+    sha = :crypto.hash_update(sha, actor_id)
+    sha = :crypto.hash_update(sha, contract_id)
+    sha = :crypto.hash_update(sha, link_name)
```
connorsmith256, line 175: What does `hash_update` do? I can't tell from the docs https://www.erlang.org/doc/man/crypto.html#hash_update-2

On the rust side it looks like we're appending bytes to a vec: https://github.com/wasmCloud/control-interface-client/pull/42/files#diff-29bfb7bb525dec2d4fc3892e18fa71485eea15e2ef2b882631e2900d76d5db81R104

autodidaddict: it's roughly the same thing. it is a "streaming hash". It's designed to allow you to have an interface where you can hash small amounts of bytes over time (like reading a giant file) so you don't have to hash the whole thing at once.

As has been verified in the tests, the two now produce the same hash.

connorsmith256: The test in the control-interface-client is starting with a sha that the comment says came from `:crypto.hash_final(sha)`, but doesn't say where the intermediate sha came from. Just wanted to double-check that the two methods always produce the same result

autodidaddict: yep, they do. per your suggestion I added a reference from one to the other.

connorsmith256 (approved):

thomastaylor312 (approved):

PR #523, opened 2023-01-04
feat: wasmtime with wasmex

ricochet:
This change moves us from wasmer to wasmtime. I am working on adding additional WASI tests but need to work through updating wascap for washlib before those work just yet.


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -183,7 +183,8 @@ defmodule HostCore.Actors.ActorModule do
       instance_id = Agent.get(agent, fn content -> content.instance_id end)
       Tracer.set_attribute("instance_id", instance_id)
 
-      {:ok, module} = Wasmex.Module.compile(bytes)
+      {:ok, store} = Wasmex.Store.new()
```
autodidaddict, line 186: This could be the diff confusing me, but doesn't this create a "non-WASI" store that won't supply the right wasi imports to the module? Wasn't the plan to always create a wasi store because it would be harmless to do so for a non-WASI actor?

ricochet: Good catch. We swapped this in one place yesterday but not the other. TY!

autodidaddict: Hah! Was just checking against my senility

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -183,7 +183,20 @@ defmodule HostCore.Actors.ActorModule do
       instance_id = Agent.get(agent, fn content -> content.instance_id end)
       Tracer.set_attribute("instance_id", instance_id)
 
-      {:ok, module} = Wasmex.Module.compile(bytes)
+      {:ok, stdin} = Wasmex.Pipe.new()
```
tessi, line 186: 👋 hey :) Thanks again for letting me know of your PR Bailey! I just want to let you know that with wasmex 0.8 one needs to set the read/write position of pipes to read them after they have been written to (e.g. to read stdout after the WASM program wrote to it, do `Wasmex.Pipe.seek(stdout, 0)`).

I don't see any use of pipes in this diff apart from creating and feeding them into wasi options, but wanted to let you know just in case :)

tessi: ohhh, I just saw that TODO further down `in the future, poll these so we can forward the err/out pipes to our logger` so probably this is no problem at all for you :)

host_core/lib/host_core/web_assembly/imports.ex
```diff
@@ -599,7 +599,8 @@ defmodule HostCore.WebAssembly.Imports do
   # Load the guest response indicated by the location and length into the :guest_response state field.
   defp guest_response(_api_type, context, agent, ptr, len) do
     memory = context.memory
-    gr = Wasmex.Memory.read_binary(memory, ptr, len)
+    caller = context.caller
+    gr = Wasmex.Memory.read_binary(caller, memory, ptr, len)
```
connorsmith256, line 603: Nit: save a couple lines with `context.caller` and `context.memory` as you've done above?

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -192,24 +205,23 @@ defmodule HostCore.Actors.ActorModule do
 
       # TODO - in the future, poll these so we can forward the err/out pipes
       # to our logger
-      {:ok, stdin} = Wasmex.Pipe.create()
-      {:ok, stdout} = Wasmex.Pipe.create()
-      {:ok, stderr} = Wasmex.Pipe.create()
+      {:ok, stdin} = Wasmex.Pipe.new()
+      {:ok, stdout} = Wasmex.Pipe.new()
+      {:ok, stderr} = Wasmex.Pipe.new()
```
connorsmith256, line 210: These look duplicated now? Aren't we already creating these pipes on line 186?

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -641,39 +653,35 @@ defmodule HostCore.Actors.ActorModule do
...
 
-    opts =
-      if imports_wasi?(Wasmex.Module.imports(module)) do
-        %{module: module, imports: imports, wasi: wasi}
-      else
-        %{module: module, imports: imports}
+    context =
+      case :ets.lookup(:module_cache, key) do
+        [{_, cached_mod}] ->
+          cached_mod
```
connorsmith256, line 673: Nit: Should this be `cached_ctx` now?

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -641,39 +653,35 @@ defmodule HostCore.Actors.ActorModule do
...
+
+        [] ->
+          {:ok, store} = Wasmex.Store.new_wasi(wasi_opts)
+          {:ok, mod} = Wasmex.Module.compile(store, bytes)
+          ctx = %{store: store, module: mod}
```
connorsmith256, line 678: Not sure I follow this: does this mean that if I start a second instance of an actor, they'll share the same WASI store/pipes?

host_core/lib/host_core/actors/actor_module.ex
ricochet: Yes, we definitely have more work to do for full wasi support 😅

host_core/lib/host_core/actors/actor_module.ex
ricochet: The store is always given it's own Wasmex.Pipe's. A second instance should mean its own pipes. Once I can actually build and sign WASI actors... I'll add some tests around this. Right now the pipes aren't used and this shouldn't be a change in behavior from our current host.

host_core/lib/host_core/actors/actor_module.ex
ricochet: Doof, thank you. I was going too fast to get this up before the community call >.<

host_core/lib/host_core/actors/actor_module.ex
connorsmith256: I might be missing something, but the way I read this, if the cache has a matching entry `(store, module)`, we use _both_ the cached store and module. And the store is just one set of pipes, right? I think we want to cache the bytes, but not the store/module now(?)

tessi: Hope it's OK if I chime in here: A store is more than a bunch of pipes. It internally holds the physical memory for all WASM structures. So `Wasmex.Memory`, `Wasmex.Instance`, and all others are fancy pointers to structures within the Store.

It's a good idea to cache the `store` and `module` together because we need both, the datastore and the pointer to a specific memorry datum in that store, to do wasmex function calls. :)

[The Wasmex docs](https://hexdocs.pm/wasmex/Wasmex.Store.html) as well as the [wasmtime docs](https://github.com/bytecodealliance/wasmtime/blob/d5b8da6eeafc3c99d0404ca764349ec691c72062/crates/wasmtime/src/store.rs#L1) may have more pointers if you want to get into the details.

brooksmtownsend (commented):
The code looks solid to me! Interested in the other conversations so I'll leave the approve off for now but I don't see anything that I'd like changed. Wonder how performance changes after this!

host_core/lib/host_core/actors/actor_module.ex
connorsmith256: Sorry for the confusion. When I said "just one" I meant it included "one set" of pipes, not that the store was _only_ composed of pipes :)

My concern here is if a user tries to start a second instance of an actor: if we match on the cache key, we will reuse the `module` and `store` from the _previous_ instance of the actor, including its pipes. I don't think that's correct: shouldn't each instance of an actor have its own set of pipes and memory? 

Today I don't think this is an issue, since wasmCloud doesn't use the module's pipes to communicate. But we're considering doing that in the future.

tessi: > When I said "just one" I meant it included "one set" of pipes, not that the store was only composed of pipes :)

ohhh, yes. sorry i misunderstood you  💡 I'm not too sure about wasmClouds architecture/nomenclature. I understand that one actor is one WASM binary, right? One actor/WASM binary could run many times in parallel.

In theory you could compile and instantiate the WASM binaries in the same `Store` meaning they'd share their IO pipes. But that doesn't sound good as instances would "fight" over these pipes. So, @connorsmith256 , you're right :) Each actor should be in their own store with their own IO pipes. Meaning, they should also have their own Memory struct (can't share them between stores) - so no re-using modules either.

One optimization could be to use `Wasmex.Engine.precompile_module()` to compile a module only once and hydrate it for every new actor (in a fresh store).

tessi:
> Wonder how performance changes after this!

do you have metrics at wasmCloud for this? I'd love to see, if I'm allowed to :)

Regarding speed, with the next release there will be a new `EngineConfig` to [configure the compilers optimization level](https://github.com/tessi/wasmex/blob/62e6980f563c3fc25a6b86d088c05daf4f5d03a1/lib/wasmex/engine_config.ex#L53). This give you the option to trade compilation speed vs. execution speed.

host_core/lib/host_core/actors/actor_module.ex
brooksmtownsend: > One optimization could be to use Wasmex.Engine.precompile_module() to compile a module only once and hydrate it for every new actor (in a fresh store).

This sounds exactly like what we want, a precompiled module that we cache the reference to and instantiate instances from.

brooksmtownsend:
> do you have metrics at wasmCloud for this? I'd love to see, if I'm allowed to :)

Yes! Sort of 😄 

I added some rudimentary performance indicators as a part of #440 so we could at least tell how fast we can invoke an actor module when directly calling the module and when sending a message over NATS (which is the way our RPC actually happens)

As a part of #465 we saw with these small tests that `wasm32-wasi` modules are actually perform about 4x better than `wasm32-unknown-unknown` modules, which was surprising! But once this PR has precompiled modules and things like that, I expect wasmtime to improve this performance even more. Worth noting that these benchmark tests really just measure the ability to call a Wasm module and return as fast as possible, it doesn't really measure any complex computations where the time to process is the most important thing. Would love to add that though, benchee certainly has the functionality for it.

brooksmtownsend:
@ricochet looks like the tests consistently fail here, are you able to reproduce the `actor_count` finding a nil list error locally?

PR #524, opened 2023-01-04
Bumping wascap version to 0.9.1

autodidaddict:
This bumps to wascap 0.9.1, which is required to support newer webassembly modules (e.g. tinygo and WASI+) signed with newer wascaps.

I had some generics issues with name overlap that led me to have to copy and paste about 7 lines of code. I'm not too worried about the duplication, just irked by the fact that it was necessary to begin with.

If any of the reviewers can figure out how to get the same `impl From` to work the way it used to, that'd be helpful.

### DIFF
--- a/host_core/native/hostcore_wasmcloud_native/src/par.rs
+++ b/host_core/native/hostcore_wasmcloud_native/src/par.rs
@@ -1,5 +1,4 @@
-use crate::{Claims, ProviderArchiveResource};
-
+use crate::{revision_or_iat, stamp_to_human, ProviderArchiveResource};
 use provider_archive::ProviderArchive;
 use rustler::{Env, Error};
 use std::env::temp_dir;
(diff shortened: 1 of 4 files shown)


host_core/native/hostcore_wasmcloud_native/src/par.rs
```diff
@@ -23,9 +22,24 @@ pub(crate) fn get_vendor(par: &ProviderArchive) -> Result<String, Error> {
     }
 }
 
-pub(crate) fn extract_claims(par: &ProviderArchive) -> Result<Claims, Error> {
+pub(crate) fn extract_claims(par: &ProviderArchive) -> Result<crate::Claims, Error> {
     par.claims()
-        .map(Claims::from)
+        .map(|c| {
```
brooksmtownsend, line 27: so this bit of copy-paste didn't work if you tried to call the function `provider_claims_to_crate_claims`?

autodidaddict: nope. still complains about type mismatch

thomastaylor312: What was the exact error? I think I can help with this

autodidaddict: It was complaining that there was no implementation of `From<...>` for the thing, and then told me that there _was_ an implementation called the same thing.

The whole thing felt very much like it was caused by overlapping or conflicting names, but I could be wrong.

autodidaddict:
Windoooooows 😠

ricochet (approved):

brooksmtownsend (approved):

Issue #527, opened 2023-01-13
[BUG] Structured log level has no effect when running a "production" host


connorsmith256:
**Describe the bug**

Despite setting `WASMCLOUD_STRUCTURED_LOG_LEVEL` to `debug`, when running a "production" build of wasmcloud-otp -- e.g. the docker image -- `Logger.debug` messages don't get printed

**To Reproduce**

1. Run `MIX_ENV=prod SECRET_KEY_BASE=teehee make run-interactive` inside `wasmcloud_host`
2. Run `require Logger`
3. Run `Logger.debug("debug message")`
4. Notice no debug message is printed
5. Run `Logger.info("info message")`
6. Notice the info message *is* printed

**Expected behavior**

The log level env var should be respected :)

**Environment (please complete the following information)**
 - OS: MacOS
 - Arch: aarch64
 - wasmCloud Version: main (v0.60-rc.1)

PR #528, opened 2023-01-13
override primary logger level when structured_log_level is set

connorsmith256:
This fixes https://github.com/wasmCloud/wasmcloud-otp/issues/527

Credit to @brooksmtownsend for the idea of just using get + Map.put to override a single key

### DIFF
--- a/host_core/lib/application.ex
+++ b/host_core/lib/application.ex
@@ -45,6 +45,11 @@ defmodule HostCore.Application do
     started = Supervisor.start_link(children, opts)
 
     if config.enable_structured_logging do
+      :logger.set_primary_config(
+        :logger.get_primary_config()
+        |> Map.put(:level, config.structured_log_level)
+      )
+
       :logger.add_handler(:structured_logger, :logger_std_h, %{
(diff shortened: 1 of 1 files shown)


host_core/lib/application.ex
```diff
@@ -45,6 +45,11 @@ defmodule HostCore.Application do
     started = Supervisor.start_link(children, opts)
 
     if config.enable_structured_logging do
+      :logger.set_primary_config(
+        :logger.get_primary_config()
+        |> Map.put(:level, config.structured_log_level)
+      )
```
brooksmtownsend, line 51: A little annoying that this is required but I accept it, looks like https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/config/config.exs#L20 is something set at _release time_ not at _run time_, right?

brooksmtownsend (approved):

host_core/lib/application.ex
connorsmith256: I think that's correct, based off this: https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#configuration

PR #530, opened 2023-01-17
Avoiding an inbox deadlock on heartbeats

autodidaddict:
There is a very specific edge case where if a heartbeat attempts to generate actor and provider data and either the actor module or the provider module are currently in the middle of processing something that can take a long time, or worse, when a ping is requested it emits an ad hoc heartbeat at or near the time of a regular interval heartbeat .... we can create a situation where calling `:get_instance_id` or `:get_link_name` etc on the OTP process can time out.

We've already indicated that when host contents get extremely large, the "old" format of the heartbeat data expands too much. There was even a `TODO` in the code to change this. By changing it so heartbeats don't actually make a call (which drops a message into an OTP process inbox) to each actor process and each provider process during a heartbeat, we now include actor count in a map and simply include a list of providers. 

Anything downstream that consumes the previous heartbeat data format will need to be modified accordingly.

The `actors` field is now a map from string to int (pk -> count), and the `providers` field is now a simple list of objects with the `link_name` and `public_key` fields.

```json
{
  "data": {
    "actors": {
      "MB2ZQB6ROOMAYBO4ZCTFYWN7YIVBWA3MTKZYAQKJMTIHE2ELLRW2E3ZW": 10
    },
    "friendly_name": "wandering-meadow-5880",
    "labels": {
      "hostcore.arch": "aarch64",
      "hostcore.os": "macos",
      "hostcore.osfamily": "unix"
    },
    "providers": [
      {
        "link_name": "default",
        "public_key": "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M"
      }
    ],
    "uptime_human": "1 minute, 32 seconds",
    "uptime_seconds": 92,
    "version": "0.60.0"
  },
  "datacontenttype": "application/json",
  "id": "d6737096-d572-400f-9242-d319ee54b8cf",
  "source": "NA2JERGCC6YGFOG662T3JYSWCDYWIJ2BBKRFJIIMULX3FK44K775TYIJ",
  "specversion": "1.0",
  "time": "2023-01-17T20:48:18.037594Z",
  "type": "com.wasmcloud.lattice.host_heartbeat"
}
```

### DIFF
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -271,8 +271,10 @@ defmodule HostCore.Actors.ActorSupervisor do
 
   @doc """
   A slightly different version of the all actors list, formatted for
-  suitability on emitted heartbeats
+  suitability on emitted heartbeats. Maps the public key of the actor
+  to the count of instances
   """
+  @spec all_actors_for_hb(host_id :: String.t()) :: %{String.t() => Integer.t()}
   def all_actors_for_hb(host_id) do
     # $1 - pk
     # $2 - pid
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -285,12 +285,14 @@ defmodule HostCore.Actors.ActorSupervisor do
         {{:"$1", :"$2", :"$3"}, [{:==, :"$3", host_id}], [{{:"$1", :"$2"}}]}
       ])
 
-    Enum.map(actors_on_host, fn {pk, pid} ->
-      {
-        pk,
-        ActorModule.instance_id(pid)
-      }
-    end)
+    Enum.reduce(actors_on_host, %{}, fn {pk, _pid}, acc -> Map.update(acc, pk, 1, &(&1 + 1)) end)
+
+    # Enum.map(actors_on_host, fn {pk, pid} ->
+    #   {
+    #     pk, "a"
+    #     #ActorModule.instance_id(pid)
+    #   }
+    # end)
   end
 
   @doc """
--- a/host_core/lib/host_core/actors/actor_supervisor.ex
+++ b/host_core/lib/host_core/actors/actor_supervisor.ex
@@ -286,13 +286,6 @@ defmodule HostCore.Actors.ActorSupervisor do
       ])
 
     Enum.reduce(actors_on_host, %{}, fn {pk, _pid}, acc -> Map.update(acc, pk, 1, &(&1 + 1)) end)
-
-    # Enum.map(actors_on_host, fn {pk, pid} ->
-    #   {
-    #     pk, "a"
-    #     #ActorModule.instance_id(pid)
-    #   }
-    # end)
   end
 
   @doc """
--- a/host_core/lib/host_core/providers/provider_supervisor.ex
+++ b/host_core/lib/host_core/providers/provider_supervisor.ex
@@ -346,6 +346,15 @@ defmodule HostCore.Providers.ProviderSupervisor do
     end)
   end
 
+  @doc """
+  Produces a list of maps, one for reach running provider on the host
+  """
+  @spec all_providers_for_hb(host_id :: String.t()) :: [
+          %{
+            required(:public_key) => String.t(),
+            required(:link_name) => String.t()
+          }
+        ]
   def all_providers_for_hb(host_id) do
     providers_on_host = providers_on_host(host_id)
 
--- a/host_core/lib/host_core/providers/provider_supervisor.ex
+++ b/host_core/lib/host_core/providers/provider_supervisor.ex
@@ -349,6 +349,17 @@ defmodule HostCore.Providers.ProviderSupervisor do
     end)
   end
 
+  def all_providers_for_hb(host_id) do
+    providers_on_host = providers_on_host(host_id)
+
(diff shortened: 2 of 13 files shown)


connorsmith256 (commented):
Mind removing the commented old code?

host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -285,12 +285,14 @@ defmodule HostCore.Actors.ActorSupervisor do
...
-      }
-    end)
+    Enum.reduce(actors_on_host, %{}, fn {pk, _pid}, acc -> Map.update(acc, pk, 1, &(&1 + 1)) end)
+
+    # Enum.map(actors_on_host, fn {pk, pid} ->
+    #   {
+    #     pk, "a"
+    #     #ActorModule.instance_id(pid)
+    #   }
+    # end)
```
connorsmith256, line 295: Still have old code here :)

connorsmith256 (dismissed):

autodidaddict:
Modified this so that it now emits a backwards-compatible heartbeat as well as the new one (it doesn't cost any extra CPU). The backwards compatible heartbeat will be retired in a future version (e.g. it's a deprecated heartbeat).

```
[#2] Received on "wasmbus.evt.default"
{"data":{"actors":[],"friendly_name":"bouncing-water-7264","labels":{"hostcore.arch":"aarch64","hostcore.os":"macos","hostcore.osfamily":"unix"},"providers":[],"uptime_human":"37 seconds","uptime_seconds":37,"version":"0.60.0"},"datacontenttype":"application/json","id":"0815f259-264a-4860-94ba-cb37154c056c","source":"NAQ6ZCLSJ4N4RD2QO2QTXLT4NVWOF6CKX7L2ZB6TYM26BNZ2DE7G7XW4","specversion":"1.0","time":"2023-01-25T13:04:26.587754Z","type":"com.wasmcloud.lattice.host_heartbeat"}

[#3] Received on "wasmbus.evt.default"
{"data":{"actors":{},"friendly_name":"bouncing-water-7264","labels":{"hostcore.arch":"aarch64","hostcore.os":"macos","hostcore.osfamily":"unix"},"providers":[],"uptime_human":"37 seconds","uptime_seconds":37,"version":"0.60.0"},"datacontenttype":"application/json","id":"95843b31-9d43-4d2a-a04e-ebd01f92a53c","source":"NAQ6ZCLSJ4N4RD2QO2QTXLT4NVWOF6CKX7L2ZB6TYM26BNZ2DE7G7XW4","specversion":"1.0","time":"2023-01-25T13:04:26.587844Z","type":"com.wasmcloud.lattice.host_heartbeat.v2"}
```

Consequences: anything listening to the old heartbeat will get empty values for contract_id and instance Id. To my knowledge, nothing actually uses that information right now as it is obtained from querying inventory.

autodidaddict:
> Just so we don't forget, this PR will be slated for the `0.61.0` release

I think now that we emit backwards compatible/deprecated heartbeats, we can put this in the 0.60 RC

host_core/lib/host_core/vhost/heartbeats.ex
```diff
@@ -4,26 +4,53 @@ defmodule HostCore.Vhost.Heartbeats do
...
   def generate_heartbeat(state) do
     config = state.config
-    # TODO: for large numbers of actors this heartbeat becomes prohibitively large and expensive
-    # refactor to only emit instance count
+
     actors =
       config.host_key
       |> ActorSupervisor.all_actors_for_hb()
-      |> Enum.map(fn {k, iid} -> %{public_key: k, instance_id: iid} end)
+      |> Map.new()
```
connorsmith256, line 49: Is this needed? If so, is it not needed for `providers`?

autodidaddict: providers is a list, this is a map. yes it's needed.

connorsmith256: It's not clear why `all_actors_for_hb` is a list that needs to be turned into a map, whereas `all_providers_for_hb` is a list that doesn't need to be turned into a map. Why not make `all_actors_for_hb` include the call to `Map.new`? And consider adding type specs to these functions to help readers understand what these structures are intended to be

autodidaddict:
> It's not clear why all_actors_for_hb is a list that needs to be turned into a map, whereas all_providers_for_hb is a list that doesn't need to be turned into a map
Because the actors field in the output is actually a map (from string -> int), whereas the list of providers is actually a list of structs. I'll add type annotations.

host_core/lib/host_core/providers/provider_supervisor.ex
```diff
@@ -247,12 +247,11 @@ defmodule HostCore.Providers.ProviderSupervisor do
     }
 
     with {:ok, {pid, _}} <- VirtualHost.lookup(host_id),
-         config <- VirtualHost.config(pid),
-         labels <- VirtualHost.labels(pid),
+         full_state <- VirtualHost.full_state(pid),
```
brooksmtownsend, line 250: From what I can gather, the host never calls `start_executable_provider` in a blocking context (only in a `ControlInterfaceTaskSupervisor` task) so this shouldn't block

brooksmtownsend (approved):
This looks to have really assisted with fixing the issues where the Vhost can get caught blocking on another GenServer.call

PR #531, opened 2023-01-17
move Registry.register calls to after module initialization

connorsmith256:
This is potentially a companion to https://github.com/wasmCloud/wasmcloud-otp/pull/530. Removing calls to get instance IDs might be desired anyway to reduce the size of heartbeat messages (in the case where many thousands of actors are running), but after talking with @brooksmtownsend, this seems to address the underlying issue, i.e. that we currently allow looking up a module from the registry before its `init` function has completed

To test, I ran this simple bash script:
```bash
while true; do
  wash ctl start actor ghcr.io/brooksmtownsend/wadice:0.2.0 --count 100 --skip-wait && wash ctl get hosts
done
```

On `main`, a host started via `make run` dies in ~10 seconds. With this change, the host stays up for many minutes (until eventually the host died because it was running so many actors the heartbeat payload got too big 🙂 )

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -154,12 +154,19 @@ defmodule HostCore.Actors.ActorModule do
 
     case start_actor(lattice_prefix, host_id, claims, bytes, oci, annotations) do
       {:ok, agent} ->
-        {:ok, agent}
+        {:ok, agent, {:continue, :register_actor}}
 
       {:error, _e} ->
         # Actor should stop with no adverse effects on the supervisor
         :ignore
     end
+  end
+
+  @impl true
+  def handle_continue(:register_actor, agent) do
+    agent_state = Agent.get(agent, fn contents -> contents end)
+    Registry.register(Registry.ActorRegistry, agent_state.claims.public_key, agent_state.host_id)
+    {:noreply, agent}
   end
 
   @impl GenServer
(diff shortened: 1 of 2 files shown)


autodidaddict:
Interesting edge case. Calling `register` inside an OTP process early. I think this one treats a problem in #530 while that treats a symptom (and makes heartbeats not be a performance problem). 👍 I think both of these together make for great stability improvements.

autodidaddict:
Build failure is coming from a bad argument to whereis: 

```
** (FunctionClauseError) no function clause matching in GenServer.whereis/1
    (elixir 1.13.4) lib/gen_server.ex:1195: GenServer.whereis(%HostCore.Providers.ProviderModule.State{annotations: %{}, contract_id: "wasmcloud:messaging", executable_path: "/tmp/wasmcloudcache/VADNMSIML2XGO2X4TPIONTIC55R2UUQGPPDZPAVSC2QD7E76CR77SPW7/1657907926/wasmcloud_messaging_default", healthy: false, host_id: "NDCRAX5IYVVPJ24ZJESJ3ZURMP4ZOIR53V2ATOVYZ7LLGPBBIFNHCUOK", instance_id: "18fbcfd6-dbf9-49ad-907b-30f4ba3a718b", lattice_prefix: "policyhome", link_name: "default", ociref: "wasmcloud.azurecr.io/nats_messaging:0.14.2", os_pid: 10674, os_port: #Port<0.66>, public_key: "VADNMSIML2XGO2X4TPIONTIC55R2UUQGPPDZPAVSC2QD7E76CR77SPW7", shutdown_delay: 300})
```

It looks like the state is being passed to whereis , which won't work. Whereis only takes pids or names (atoms)

thomastaylor312 (approved):

PR #535, opened 2023-01-21
fix various bugs related to policy checks

connorsmith256:
This PR fixes a few bugs that would prevent actors/providers from starting/invocations from being processed. I also refactored `start_actor` to have significantly less nesting in an attempt to make it easier to read.

I tested these changes by:
- starting a host
- starting actors on the host (including multiple counts)
- starting providers on the host
- linking actors and providers
- issuing requests that generate both actor -> provider and provider -> actor invocations

Everything seems to be working now

Signed-off-by: Connor Smith <connor.smith.256@gmail.com>

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -498,15 +498,17 @@ defmodule HostCore.Actors.ActorModule do
   defp policy_check(%{source_target: true} = token, agent) do
     lattice_prefix = Agent.get(agent, fn contents -> contents.lattice_prefix end)
     host_id = Agent.get(agent, fn contents -> contents.host_id end)
-    config = VirtualHost.config(host_id)
-    origin = token.invocation["origin"]
+    source = token.invocation["origin"]
     target = token.invocation["target"]
 
     decision =
-      with {:ok, _topic} <- PolicyManager.policy_topic(config),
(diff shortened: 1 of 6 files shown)


host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -518,18 +520,31 @@ defmodule HostCore.Actors.ActorModule do
...
+              issuedOn: source_claims[:iat],
+              expiresAt: source_claims[:exp],
+              expired: expired
+            },
+            %{
+              publicKey: target["public_key"],
+              contractId: target["contract_id"],
+              linkName: target["link_name"],
+              issuer: target_claims[:iss]
+            },
```
connorsmith256, line 541: These were mistakingly removed

host_core/lib/host_core/policy/manager.ex
```diff
@@ -71,7 +72,7 @@ defmodule HostCore.Policy.Manager do
           publicKey: host_config.host_key,
           issuer: host_config.cluster_key,
           latticeId: host_config.lattice_prefix,
-          labels: host_config.labels,
+          labels: labels,
```
connorsmith256, line 75: `host_config` does not contain a `:labels` key, causing this to silently fail every time the function is called

autodidaddict (approved):
Thou art a saint

PR #536, opened 2023-01-22
Fixes issues with deleting linkdefs

autodidaddict:
This addresses two problems:

1) This will change the ID of legacy linkdefs to the new hash-based ID so they can be properly looked up in the new host
2) This actually deletes linkdefs from the metadata cache via the control interface server

## Testing Method
Added a link (these are fake IDs)
```
wash ctl link put --link-name default MC7KRLM6QAMNTJY2S7YEMINQ27CADS2UT27EYT4CBQHLZAMGYBRYAIYD VCOX3XM5ZQTX6W7DETWQM4U7JZ7T7EJPUOHNQQX4F3JPJJWCNB6LTWSY wasmcloud:messaging
```
Made sure it shows up in the cache:
```
 nats kv ls LATTICEDATA_default
LINKDEF_13599B7EBAA569030B1B9580585B05B2980682D5717F7C9E1910F27339D47F3D
```
Now delete the link:
```
wash ctl link del MC7KRLM6QAMNTJY2S7YEMINQ27CADS2UT27EYT4CBQHLZAMGYBRYAIYD wasmcloud:messaging
```
And now verify that the link is indeed gone:
```
nats kv ls LATTICEDATA_default
No keys found in bucket
```

Restart the host and verify that it ignores (or no-op deletes) the now-purged key:
```
14:03:11.782 [debug] Removing cached reference for linkdef ID 13599B7EBAA569030B1B9580585B05B2980682D5717F7C9E1910F27339D47F3D
14:03:11.782 [warning] Attempted to remove non-existent linkdef 13599B7EBAA569030B1B9580585B05B2980682D5717F7C9E1910F27339D47F3D (this is OK)
```

It's interesting to note that if a host starts up in a lattice that has had an enormous amount of churn, it could see a few dozen no-op skip deletes at startup, but this won't grow infinitely because the hash of a linkdef will always be the same regardless of the values payload.

### DIFF
--- a/host_core/lib/host_core/jetstream/client.ex
+++ b/host_core/lib/host_core/jetstream/client.ex
@@ -2,6 +2,9 @@ defmodule HostCore.Jetstream.Client do
   @moduledoc false
   use GenServer
 
+  @kvoperation "KV-Operation"
+  @kvpurge "PURGE"
+
   alias HostCore.Jetstream.LegacyCacheLoader
 
   require Logger
@@ -185,6 +188,33 @@ defmodule HostCore.Jetstream.Client do
     end
 
     {:noreply, state}
+  end
+
+  def kv_del(lattice_prefix, "", key), do: kv_del(lattice_prefix, nil, key)
+
+  def kv_del(lattice_prefix, js_domain, key) do
+    # delete requires us to make a request on the topic with headers
+    # KV-Operation : PURGE
+    topic = kv_operation_topic(lattice_prefix, key, js_domain)
+    IO.inspect(topic)
+    headers = [{@kvoperation, @kvpurge}]
+
+    # TODO: once the Jetstream hex package support js_domains, switch this
+    # code to use JetStream.API.xxxx
+    case lattice_prefix
+         |> HostCore.Nats.control_connection()
+         |> Gnat.request(topic, <<>>, headers: headers) do
+      {:ok, _} ->
+        Logger.debug("Deleted key #{key} from metadata lattice bucket #{lattice_prefix}")
+        :ok
+
+      {:error, e} ->
+        Logger.error(
+          "Failed to delete key #{key} from metadata bucket #{lattice_prefix}: #{inspect(e)}"
+        )
+
+        {:error, e}
+    end
   end
 
   # Make sure that where applicable `value` is already encoded as JSON because this function won't do it.
--- a/host_core/lib/host_core/jetstream/client.ex
+++ b/host_core/lib/host_core/jetstream/client.ex
@@ -196,7 +196,6 @@ defmodule HostCore.Jetstream.Client do
     # delete requires us to make a request on the topic with headers
     # KV-Operation : PURGE
     topic = kv_operation_topic(lattice_prefix, key, js_domain)
-    IO.inspect(topic)
     headers = [{@kvoperation, @kvpurge}]
 
     # TODO: once the Jetstream hex package support js_domains, switch this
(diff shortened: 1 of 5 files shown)


host_core/lib/host_core/linkdefs/manager.ex
```diff
@@ -62,7 +66,8 @@ defmodule HostCore.Linkdefs.Manager do
     :ets.insert(table_atom(lattice_prefix), {ld.id, ld})
   end
 
-  def uncache_link_definition(lattice_prefix, ldid) do
+  def uncache_link_definition(lattice_prefix, ldid, js_domain) do
+    HostCore.Jetstream.Client.kv_del(lattice_prefix, js_domain, "LINKDEF_#{ldid}")
```
connorsmith256, line 70: This doesn't feel right, based off the name of this function. It feels like both `uncache_link_definition` and `kv_del` should be called together, in most cases, but removing from an in-memory cache and removing from the source of truth feel like two different operations

host_core/lib/host_core/jetstream/client.ex
```diff
@@ -187,6 +190,33 @@ defmodule HostCore.Jetstream.Client do
     {:noreply, state}
   end
 
+  def kv_del(lattice_prefix, "", key), do: kv_del(lattice_prefix, nil, key)
+
+  def kv_del(lattice_prefix, js_domain, key) do
+    # delete requires us to make a request on the topic with headers
+    # KV-Operation : PURGE
+    topic = kv_operation_topic(lattice_prefix, key, js_domain)
+    IO.inspect(topic)
```
connorsmith256, line 199: Nit: remove me

host_core/lib/host_core/linkdefs/manager.ex
connorsmith256: I'm curious about these two flows now:

With control interface client prior to v0.23:
- `wash ctl link del....`
- request issued to `wasmbus.ctl.[lattice].linkdefs.del`
- handled by `del_link_definition_by_triple`
- `del_link_definition` called
- `kv_del` called, which publishes to KV topic
- triggers a `handle_action` in `MetadataCacheLoader` (?)
- `del_link_definition` called (again)
- `kv_del` called (again), which publishes to KV topic again (no-op?)

With control interface client v0.23:
- `wash ctl link del...`
- deletes from KV directly
- triggers a `handle_action` in `MetadataCacheLoader`
- `del_link_definition` called
- `kv_del` called, which publishes to KV topic, but the key is already deleted, so no-op

Assuming I've followed the flow correctly, neither of these feel quite right. I think the flows should be:

Responding to a request on `request to `wasmbus.ctl.[lattice].linkdefs.del` (which is only hit by deprecated clients):
- simply call `kv_del`, deleting the key from the bucket
- (the `MetadataCacheLoader` should then get and handle a `:key_deleted` action, as below)

Responding to `:key_deleted` actions:
- uncache link def from ETS table
- broadcast event to other hosts in lattice

connorsmith256 (changes requested):

host_core/lib/host_core/linkdefs/manager.ex
autodidaddict: The motivation behind deleting from the in-memory cache and calling the KV del is to handle the off chance that the key deleted message doesn't make its way to the host.

autodidaddict: > (the MetadataCacheLoader should then get and handle a :key_deleted action, as below)
This isn't 100% guaranteed. I'd rather delete it immediately especially since there's really no downside.

autodidaddict: >broadcast event to other hosts in lattice
This shouldn't be done because all other hosts in the lattice will have already received the key deleted message from the bucket subscription.

connorsmith256 (approved):
Okay, the new code makes sense and LGTM 👍

PR #539, opened 2023-01-26
Putting contract ID back on heartbeat

autodidaddict:
This puts the capability provider's contract ID into the claims cache, which in turn stores it in the kv bucket. Once this data is in the claims cache, we can now query it and put it back on the heartbeat without performing a blocking GenServer call

### DIFF
--- a/host_core/lib/host_core/claims/manager.ex
+++ b/host_core/lib/host_core/claims/manager.ex
@@ -84,7 +84,7 @@ defmodule HostCore.Claims.Manager do
         end,
       version: claims.version,
       sub: claims.public_key,
-      contract_id: Map.get(claims, :contract_id)
+      contract_id: Map.get(claims, :contract_id) || ""
     }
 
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/claims/manager.ex
```diff
@@ -82,7 +83,8 @@ defmodule HostCore.Claims.Manager do
           Enum.join(claims.tags, ",")
         end,
       version: claims.version,
-      sub: claims.public_key
+      sub: claims.public_key,
+      contract_id: Map.get(claims, :contract_id)
```
brooksmtownsend, line 87: If this is an actor, we shouldn't add the contract ID here (or at least it should be "n/a") since making it null can cause deserialization errors down the line

brooksmtownsend (changes requested):
Contract ID being null here violates some serialization down the line and we should take care not to put it in (or put N/a) for actors

brooksmtownsend (approved):

Issue #502, opened 2022-11-21
[BUG] Hot reloading fails to create hotwatch


brooksmtownsend:
**Describe the bug**

When attempting to hotwatch a file on disk, I get this message:
```
16:52:07.234 [info] Starting actor MAVU7CIQWXSTM5AW6OGSVGQMDNTFUOFI7RPOG435O4G7Z75Q2WXEA6DI
16:52:07.294 [error] GenServer :actor_watcher terminating
** (MatchError) no match of right hand side value: :ignore
    (wasmcloud_host 0.57.1) lib/wasmcloud_host/actor_watcher.ex:91: WasmcloudHost.ActorWatcher.handle_call/3
    (stdlib 4.0.1) gen_server.erl:1146: :gen_server.try_handle_call/4
    (stdlib 4.0.1) gen_server.erl:1175: :gen_server.handle_msg/6
    (stdlib 4.0.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
Last message (from #PID<0.2522.0>): {:hotwatch_actor, "/Users/brooks/github.com/wasmcloud/examples/actor/echo-messaging/build/echo_messaging_s.wasm", 1}
16:52:07.295 [error] GenServer #PID<0.2522.0> terminating
** (stop) exited in: GenServer.call(:actor_watcher, {:hotwatch_actor, "/Users/brooks/github.com/wasmcloud/examples/actor/echo-messaging/build/echo_messaging_s.wasm", 1}, 60000)
    ** (EXIT) an exception was raised:
        ** (MatchError) no match of right hand side value: :ignore
            (wasmcloud_host 0.57.1) lib/wasmcloud_host/actor_watcher.ex:91: WasmcloudHost.ActorWatcher.handle_call/3
            (stdlib 4.0.1) gen_server.erl:1146: :gen_server.try_handle_call/4
            (stdlib 4.0.1) gen_server.erl:1175: :gen_server.handle_msg/6
            (stdlib 4.0.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
    (elixir 1.13.4) lib/gen_server.ex:1030: GenServer.call/3
    (wasmcloud_host 0.57.1) lib/wasmcloud_host_web/live/components/start_actor_component.ex:69: StartActorComponent.handle_event/3
[...]

PR #543, opened 2023-02-03
reference updated function

connorsmith256:
This PR removes a reference to a function that doesn't exist anymore. On main, trying to load metrics results in an internal server error. This references the replacement function, allowing the page to load again

I'm unfamiliar with the metrics page, so I can't attest this is now showing the correct information, but I can say the page loads now

Signed-off-by: Connor Smith <connor.smith.256@gmail.com>


brooksmtownsend (approved):

brooksmtownsend:
This kind of thing is likely what's causing the issues in #502, though we'll tackle that in the future

PR #544, opened 2023-02-07
Cache call alias when claims are replicated

brooksmtownsend:
This PR ensures that when we receive claims over jetstream we attempt to cache call aliases as well for invocation lookups. The symptom that I ran into was two actors running on two wasmCloud hosts, and one host couldn't find the call alias for the other.

Just as verification, and I know these logs are a handful and a half, I put the petclinic API and httpserver on one host, and the petclinic UI (which is invoked via call alias) on a second host. Works great 👍🏻 
image

### DIFF
--- a/host_core/lib/host_core/claims/manager.ex
+++ b/host_core/lib/host_core/claims/manager.ex
@@ -30,15 +30,16 @@ defmodule HostCore.Claims.Manager do
 
   @spec cache_claims(
           lattice_prefix :: String.t(),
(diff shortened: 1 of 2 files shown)


autodidaddict (dismissed):

autodidaddict (approved):
LGTM because the `cache_call_alias` function is nil-safe

connorsmith256 (approved):
LGTM

PR #558, opened 2023-03-07
Pass expected name for enabling structured logging in providers

connorsmith256:
## Feature or Problem
Currently, the host passes `enable_structured_logging` as a field in the `HostData` given to providers via stdin when they start. However, the [smithy definition](https://github.com/wasmCloud/interfaces/blob/0fe4ba01cccbe2342a987f01586efa8599557458/core/wasmcloud-core.smithy#L180) of `HostData` expects the field to be serialized as `structured_logging`.

In effect, there is no way to enable structured logging in providers today, since we've misspelled the field.

## Related Issues
This doesn't fix https://github.com/wasmCloud/weld/issues/140, but it is related

## Release Information
Next release

## Consumer Impact
The impact should be positive, since we're enabling structured logs for providers, unifying the log format between the host and providers. However, on the off chance that users are relying on this buggy behavior today, this could _technically_ be considered a breaking change(?)

### Unit Test(s)
N/A

### Acceptance or Integration
N/A

### Manual Verification
- started a host with `WASMCLOUD_STRUCTURED_LOGGING_ENABLED=true WASMCLOUD_STRUCTURED_LOG_LEVEL=debug make run`
- started the HTTP server provider, echo actor, and linked them
- made a request and confirmed the provider emitted structured log messages like `{"timestamp":"2023-03-07T16:44:24.744883Z","level":"INFO","fields":{"message":"finished processing with success","status":200},"target":"warp::filters::trace"}`


brooksmtownsend (dismissed):

thomastaylor312 (approved):
This is a breaking change, although small. So you were thinking right

autodidaddict (approved):

brooksmtownsend (approved):

PR #559, opened 2023-03-07
unpin from specific SHA of bindle

connorsmith256:
## Feature or Problem
We've been pinned to a specific SHA of bindle because at the time v0.9 hadn't been released. It has now

## Related Issues
N/A

## Release Information
Next

## Consumer Impact
N/A

### Manual Verification
I'll be honest, I didn't test this. Besides compiling, what should be done?


thomastaylor312 (dismissed):
I don’t remember if we have a lock file, but if we do can you make sure it is updated?

brooksmtownsend (approved):
Same comment as Taylor

brooksmtownsend (approved):

Issue #519, opened 2022-12-15
Add new cloud event for provider health check status change


autodidaddict:
Right now there's no way to remotely monitor whether a given provider has had its health check state go from pass to fail or from fail to pass. To avoid spamming, we shouldn't emit duplicate events in succession if the provider is in the same state as the last health check.

When a provider switches from passing a health check to failing, it should emit a `provider_health_check_failed` cloud event on the `wasmbus.evt.{x}` topic. Conversely, if a provider passes a health check if the previous status was a fail, then it should emit a `provider_health_check_passed` cloud event.

The `ProviderModule` `GenServer` is probably a good place to stick a `last_check` result state field that can be used to compare after performing the RPC-based health check over NATS.

This issue stems from a discussion with @jordan-rash , CC @giusdp .

jordan-rash:
The Healthcheck contract has two fields.  If we are only emitting changes (true/false) in the `healthy` boolean, we are potentially missing a lot of good context in the `message`.  For example, I could have something thats healthy, but some state is deteriorated.  I don't want to mark the asset unhealthy, but I do what to be notified.  Currently, we request health every 30 seconds, I dont think it would be too noisy to just emit the status with message every request with a event of `provider_health_check_status` or something similar????

```smithy
/// Return value from actors and providers for health check status
@wasmbusData
structure HealthCheckResponse {
  /// A flag that indicates the the actor is healthy
  @n(0)
  healthy: Boolean
  
  /// A message containing additional information about the actors health
  @n(1)
  message: String
}
```

autodidaddict:
It's 30 seconds per running provider. That's probably not that bad, considering the host emits its own heartbeat every 30 seconds and there's rarely more than 10 providers on a host

giusdp:
I'm slowly learning the host_core codebase. I saw the providers are GenServers (managed by a provider supervisor) and they handle a `:do_health` message. I saw this snippet in that function:

 ```elixir
   # Only publish health check pass/fail when state changes
    state =
      case res do
        {:ok, _body} when not state.healthy ->
          publish_health_passed(state)
          %State{state | healthy: true}

        {:ok, _body} ->
          state

        {:error, _} ->
          if state.healthy do
            publish_health_failed(state)
            %State{state | healthy: false}
          else
            state
          end
      end  
```

Is it not already sending the health updates **only** when state actually changes?

jordan-rash:
[...]

Issue #517, opened 2022-12-13
[FEATURE] Change Inbox Prefix for Healthchecks


jordan-rash:
**Is your feature request related to a problem? Please describe.**
I would like to be able to pull out the healthcheck responses, but currently, they are random `_INBOX.uuid` topic names

**Describe the solution you'd like**
I would like to utilize the [custom inbox prefix](https://docs.rs/async-nats/latest/async_nats/struct.ConnectOptions.html#method.custom_inbox_prefix) functionality to make it easier to subscribe to just healthcheck responses.

**Describe alternatives you've considered**
I have tried to script out listening to the requests and then picking up the responses...i didn't like it.

I am happy to take a wack at this if someone can point me at the right section of code and there is a consensus on what the prefix should be

jordan-rash:
Pictures!

<img width="947" alt="Screenshot 2022-12-13 at 7 36 45 AM" src="https://user-images.githubusercontent.com/15827604/207362222-af6c0c4f-145c-4626-b1e1-886a55bfa10a.png">

brooksmtownsend:
🤔 Aren't these inbox prefixes just used when something else does a `nats req`? And for your use case, you're just looking to hook into each health check so you can monitor the health of the components, right?

autodidaddict:
>I would like to be able to pull out the healthcheck responses, but currently, they are random _INBOX.uuid topic names

Snooping on other requester's `_INBOX` topics is in most cases, a security violation for a number of reasons. In most production scenarios, users are configured to be unable to listen on other people's reply topics. If you want to monitor the health of things from somewhere else, you can watch the events on the `wasmbus.evt.xxx` topic. 

If you think additional events should be published on that stream, then I think that's probably a better place to put what you need to observe.

autodidaddict:
Also, as Brooks indicated, the custom inbox prefix in the Rust crate (and Go library) are only for setting the reply-to field when making requests, not for listening to responses from other user's requests.

jordan-rash:
I acknowledge it wasn't the intended use of the prefix, but I guess I was trying to identify a way to hook into the health check message.

From monitoring the nats traffic via `nats sub ">"` (which should grab everything), after the initial `health_check_passed` message, I don't see anything from the providers in `wasmbus.evt....`, am I missing something?

autodidaddict:
This is what I was getting at. We could make it so that the host itself emits an event when a provider changes health status. It doesn't do this today

[...]

PR #560, opened 2023-03-08
Add provider health status event

giusdp:
## Feature or Problem
From the discussions of the following issues, providers should emit the health events on the `wasmbus.evt` topic.
Please let me know if I misunderstood what was being asked in those issues or if there is something missing.

## Related Issues
- #517 
- #519 

## Release Information
Next

## Consumer Impact
N/A

## Testing

Built on platform(s)
- [x] x86_64-linux

Tested on platform(s)
- [x] x86_64-linux

### Unit Test(s)
I've added a test in test/host_core/provider_tests.exs to start a provider from file and wait on the health_check events. It first waits for the health_check_passed event when the provider is started and afterwards waits for the health_check_status event once, which is emitted every 30 seconds.

### Manual Verification

I've launched nats and subscribed to `wasmbus.evt.*` and to ">". 
I've used `make run` from the dashboard project and from the web page started a provider from file to checkout that the health_status event was being correctly sent.

### DIFF
--- a/host_core/lib/host_core/providers/provider_module.ex
+++ b/host_core/lib/host_core/providers/provider_module.ex
@@ -395,6 +395,7 @@ defmodule HostCore.Providers.ProviderModule do
           %State{state | healthy: true}
 
         {:ok, _body} ->
+          publish_health_status(state)
           state
 
         {:error, _} ->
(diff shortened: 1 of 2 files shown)


host_core/lib/host_core/providers/provider_module.ex
```diff
@@ -395,6 +395,7 @@ defmodule HostCore.Providers.ProviderModule do
           %State{state | healthy: true}
 
         {:ok, _body} ->
+          publish_health_status(state)
```
connorsmith256, line 398: If we're always going to publish status events, I think this block should be refactored. Something like:
```elixir
new_state = case res do
  {:ok, _body} when not state.healthy ->
    %State(state | healthy: true}
  {:error, _} when state.healthy ->
    %State(state | healthy: false}
  _ ->
    state
end
publish_health_status(new_state)
{:noreply, new_state}
```

(`publish_health_status` might need to change too, to infer what the name of the cloud event should be from whether the state is healthy or not)

This separates updating state, which we want to do when the health status changes, from publishing, which we want to do every time now

connorsmith256 (changes requested):
Thank you for another contribution @giusdp !

I have one request to make the code more readable

host_core/lib/host_core/providers/provider_module.ex
giusdp: Thanks for the suggestion! I refactored the code and changes `publish_health_status(new_state)` to handle the cases for the passed, failed health event and the normal status event.

connorsmith256 (approved):
LGTM!

connorsmith256:
I'm going to hold off merging for now since we're hoping to cut a release for v0.61 this week, but thanks again!

giusdp:
No problem!

PR #567, opened 2023-03-13
Compare OCI sha with disk cache sha before pulling an image

aish-where-ya:
## Feature or Problem
When fetching an OCI image, the code checked whether a cache existed and pulled the image if it did not exist.
Now it checks the cache file she with the latest oci image she to verify whether the latest image is cached and pulls if the she digest does not match.

## Related Issues
NA

## Release Information
next

## Consumer Impact
Users will have no breaking changes. 

## Testing
Manual testing was performed using the nif since unit tests are not supported yet. End to end tests can be added as future work.

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Manual unit tests were performed. Following is a log showing a sha digest match on 3 cache files resulting in no additional image pulls.

```
..cache file: "/var/folders/k6/7fp1nc5s7w1987jh7vzpnyn40000gn/T/wasmcloud_ocicache/wasmcloud_azurecr_io_echo_0_3_1.bin" 
oci image: Reference { registry: "wasmcloud.azurecr.io", repository: "echo", tag: Some("0.3.1"), digest: None } 
OCI digest: [109, 8, 23, 54, 89, 105, 252, 166, 166, 139, 200, 184, 80, 191, 212, 68, 48, 192, 232, 180, 191, 115, 141, 176, 201, 169, 227, 234, 80, 144, 79, 197] 
cache file sha: [109, 8, 23, 54, 89, 105, 252, 166, 166, 139, 200, 184, 80, 191, 212, 68, 48, 192, 232, 180, 191, 115, 141, 176, 201, 169, 227, 234, 80, 144, 79, 197] 
MATCH!!!!

..cache file: "/var/folders/k6/7fp1nc5s7w1987jh7vzpnyn40000gn/T/wasmcloud_ocicache/wasmcloud_azurecr_io_echo_0_3_1.bin" 
oci image: Reference { registry: "wasmcloud.azurecr.io", repository: "echo", tag: Some("0.3.1"), digest: None } 
OCI digest: [109, 8, 23, 54, 89, 105, 252, 166, 166, 139, 200, 184, 80, 191, 212, 68, 48, 192, 232, 180, 191, 115, 141, 176, 201, 169, 227, 234, 80, 144, 79, 197] 
cache file sha: [109, 8, 23, 54, 89, 105, 252, 166, 166, 139, 200, 184, 80, 191, 212, 68, 48, 192, 232, 180, 191, 115, 141, 176, 201, 169, 227, 234, 80, 144, 79, 197] 
MATCH!!!!

cache file: "/var/folders/k6/7fp1nc5s7w1987jh7vzpnyn40000gn/T/wasmcloud_ocicache/wasmcloud_azurecr_io_httpserver_0_14_0.bin" 
oci image: Reference { registry: "wasmcloud.azurecr.io", repository: "httpserver", tag: Some("0.14.0"), digest: None } 
OCI digest: [79, 231, 157, 100, 0, 141, 55, 210, 159, 204, 234, 250, 180, 55, 182, 136, 58, 57, 26, 209, 103, 109, 251, 19, 21, 29, 51, 120, 160, 23, 203, 241] 
cache file sha: [79, 231, 157, 100, 0, 141, 55, 210, 159, 204, 234, 250, 180, 55, 182, 136, 58, 57, 26, 209, 103, 109, 251, 19, 21, 29, 51, 120, 160, 23, 203, 241] 
MATCH!!!!
```

### Acceptance or Integration
NA

### Manual Verification
Manual verification described above.

### DIFF
--- a/host_core/native/hostcore_wasmcloud_native/src/oci.rs
+++ b/host_core/native/hostcore_wasmcloud_native/src/oci.rs
@@ -2,9 +2,12 @@ use std::collections::HashMap;
 use std::env::{temp_dir, var};
 use std::path::PathBuf;
 use std::str::FromStr;
-use tokio::io::AsyncWriteExt;
-
+use tokio::fs::File;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use sha2::{Sha256, Digest};
+use hex;
 use oci_distribution::secrets::RegistryAuth;
+use oci_distribution::manifest::OciManifest;
 
 pub(crate) const OCI_VAR_REGISTRY: &str = "OCI_REGISTRY";
 pub(crate) const OCI_VAR_USER: &str = "OCI_REGISTRY_USER";
@@ -50,31 +53,56 @@ pub(crate) async fn fetch_oci_path(
             "Fetching images tagged 'latest' is currently prohibited in this host. This option can be overridden with WASMCLOUD_OCI_ALLOW_LATEST".into());
     }
     let cf = cached_file(img).await?;
-    if tokio::fs::metadata(&cf).await.is_err() {
-        let auth = determine_auth(img, creds_override);
-        let img = oci_distribution::Reference::from_str(img)?;
 
-        let protocol =
-            oci_distribution::client::ClientProtocol::HttpsExcept(allowed_insecure.to_vec());
-        let config = oci_distribution::client::ClientConfig {
-            protocol,
-            ..Default::default()
-        };
-        let mut c = oci_distribution::Client::new(config);
+    let auth = determine_auth(img, creds_override);
+    let img = oci_distribution::Reference::from_str(img)?;
+
+    let protocol =
+        oci_distribution::client::ClientProtocol::HttpsExcept(allowed_insecure.to_vec());
+    let config = oci_distribution::client::ClientConfig {
+        protocol,
+        ..Default::default()
+    };
+    let mut c = oci_distribution::Client::new(config);
+
+    if tokio::fs::metadata(&cf).await.is_err() {     
         let imgdata = pull(&mut c, &img, &auth).await;
-
         match imgdata {
             Ok(imgdata) => {
-                let mut f = tokio::fs::File::create(&cf).await?;
-                let content = imgdata
-                    .layers
-                    .into_iter()
-                    .flat_map(|l| l.data)
-                    .collect::<Vec<_>>();
-                f.write_all(&content).await?;
-                f.flush().await?;
+                cache_oci_image(&imgdata, &cf).await;
             }
             Err(e) => return Err(format!("Failed to fetch OCI bytes: {}", e).into()),
+        }
+    }else{
+        let manifest = c.pull_manifest(&img, &auth).await;
+        match manifest{
+            Ok(manifest)=>{
+                let (oci_manifest, _) = manifest;
+                if let OciManifest::Image(image_manifest) = oci_manifest {
+                    
+                    let layers = image_manifest.layers;
+                    // The digest will be string like sha256:asdf, and the first 7 characters are removed
+                    let oci_sha = &layers[0].digest[7..];
+                    let oci_sha_bytes: Vec<u8> = hex::decode(oci_sha).expect("Invalid Hex String");
+
+                    let mut cache_file = File::open(cf.clone()).await?;
+                    let mut cache_contents = Vec::new();
+                    cache_file.read_to_end(&mut cache_contents).await?;
+                    let cache_sha = Sha256::digest(cache_contents);
+
+                    // compare the OCI sha with the cache file sha and pull the OCI image
+                    if oci_sha_bytes[..] != cache_sha[..]{
+                        let imgdata = pull(&mut c, &img, &auth).await;
+                        match imgdata {
+                            Ok(imgdata) => {
+                                cache_oci_image(&imgdata, &cf).await;
+                            }
+                            Err(e) => return Err(format!("Failed to fetch OCI bytes: {}", e).into()),
+                        }
+                    }
+                }            
+            }
+            Err(e) => return Err(format!("Failed to fetch OCI manifest: {}", e).into()),
         }
     }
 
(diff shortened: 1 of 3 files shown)


host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -112,3 +140,17 @@ async fn pull(
         .await
         .map_err(|e| format!("{}", e).into())
 }
+
+async fn cache_oci_image(
+    imgdata: &oci_distribution::client::ImageData,
+    cache_filename: &PathBuf,
```
connorsmith256, line 146: Nit: rename to `path` (since it's the full path, not just the file name)?

host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -112,3 +140,17 @@ async fn pull(
...
+) {
+    let mut f = tokio::fs::File::create(&cache_filename).await.unwrap();
```
connorsmith256, line 148: These unwraps aren't safe (e.g. what if the directory doesn't exist or the host doesn't have permissions to create a file in it). You'll need to handle the error cases. Since the operations here can fail, I think the function should return a `std::io::Result<()>`. That would let you use `?` instead of `unwrap` on these, since it looks like they're all tokio fs calls which return `std::io::Result` already

host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -50,32 +53,57 @@ pub(crate) async fn fetch_oci_path(
...
-                    .into_iter()
-                    .flat_map(|l| l.data)
-                    .collect::<Vec<_>>();
-                f.write_all(&content).await?;
-                f.flush().await?;
+                cache_oci_image(&imgdata, &cf).await;
             }
             Err(e) => return Err(format!("Failed to fetch OCI bytes: {}", e).into()),
         }
+    }else{
```
connorsmith256, line 76: I'm not sure if we have format checking set up in the NIF, but you'll want to run `cargo fmt -- --check && cargo clippy -- -D warnings` from the `hostcore_wasmcloud_native` directory

connorsmith256 (changes requested):

host_core/native/hostcore_wasmcloud_native/src/oci.rs
aish-where-ya: Ahh missed that one.

host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -50,32 +53,57 @@ pub(crate) async fn fetch_oci_path(
...
+        let manifest = c.pull_manifest(&img, &auth).await;
+        match manifest{
+            Ok(manifest)=>{
+                let (oci_manifest, _) = manifest;
```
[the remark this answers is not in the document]
aish-where-ya: So the reason that we cannot use this hash is because it is computed for the manifest. This may not be same as the OCI Image hash. The Oci Descriptor has multiple digest fields, but the one that we are using (inside layers) is the one that matches the cache file contents.

host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -50,32 +53,57 @@ pub(crate) async fn fetch_oci_path(
...
+                if let OciManifest::Image(image_manifest) = oci_manifest {
+                    
+                    let layers = image_manifest.layers;
+                    // The digest will be string like sha256:asdf, and the first 7 characters are removed
+                    let oci_sha = &layers[0].digest[7..];
```
thomastaylor312, line 85: Two small notes here: I'd check that the `layers` vec is not empty so we don't get a panic here. Second is a nit that I think it would read clearer if you did `layers[0].digest.trim_left_matches("sha256:")`

host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -50,32 +53,57 @@ pub(crate) async fn fetch_oci_path(
...
+                    let oci_sha_bytes: Vec<u8> = hex::decode(oci_sha).expect("Invalid Hex String");
```
thomastaylor312, line 86: Oh! One more helpful tip here, you don't actually need to decode as hex if you don't want to. When you hash the file below, you can use `format!("{:x}", hash_result)` and compare the strings. See the example here: https://github.com/deislabs/bindle/blob/f8d20c506e3abab5a5e0fa90e0940d6db84a0908/src/id.rs#L78.

No need to change anything, was just pointing it out for future sanity

host_core/native/hostcore_wasmcloud_native/src/oci.rs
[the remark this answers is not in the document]
connorsmith256: Sounds good

host_core/native/hostcore_wasmcloud_native/src/oci.rs
```diff
@@ -1,13 +1,12 @@
+use oci_distribution::manifest::OciManifest;
+use oci_distribution::secrets::RegistryAuth;
+use sha2::{Digest, Sha256};
```
connorsmith256, line 3: Nit: for imports our convention is:
```
[std lib]
// empty line
[3rd party like oci_distribution]
// empty line
[1st party like crate::foo]
```

connorsmith256 (commented):
Just one nit from me, and Taylor found another place this could panic. Once those are resolved LGTM

host_core/native/hostcore_wasmcloud_native/src/oci.rs
aish-where-ya: Yes, I saw Taylor's review comments after pushing my commit. My bad. 
The imports were something that clippy/rust fmt recommended, I will change it back

connorsmith256 (approved):

PR #568, opened 2023-03-13
Adds missing fields and updates some dependencies to the right version

thomastaylor312:
Was using the chart recently and realized we hadn't added some new config values or updated the version of NATS for a while.


brooksmtownsend (approved):

thomastaylor312:
The build failure is definitely a flake, especially as this is just changing YAML files. So I'm just going to merge

PR #569, opened 2023-03-14
Support wasm components by converting wasm functions to use wasmCloud Rust runtime

autodidaddict:
Currently stuck/floundering on the only two pieces of this that actually matter:
* reproducing the wasmex-style callback system, but instead of a list of dynamic type
* instantiating actors from the wasmcloud runtime and then using those to execute functions

submitting the draft so folks can see where I am


host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
```diff
@@ -0,0 +1,318 @@
...
+    from: Term,
+) -> rustler::Atom {
+    let pid = env.pid();
+    let mut thread_env = OwnedEnv::new();
+
+    let from = thread_env.save(from);
+    let payload = payload.to_vec();
+    let operation = operation.to_owned();
+
+    println!("HEY THERE");
```
rvolosatovs, line 172: :wave:

autodidaddict:
New instructions for playing with it: start your host with `iex -S mix` (your host ID will vary)

```
{:ok, {pid, foo}} = HostCore.Vhost.VirtualHost.lookup("NBJCCFTGKFJ7MNSQXQGVU5QRP4AIJSAXWUKYXJWK7LH3T5CJZ4FFN3LD")

{:ok, rtpid} = HostCore.Vhost.VirtualHost.get_runtime(pid)

raw = File.read!("/Users/kevin/code/wasmcloud/wasmcloud-otp/host_core/test/fixtures/actors/echo.wasm")

{:ok, aref} = HostCore.WasmCloud.Runtime.Server.precompile_actor(rtpid, raw)

HostCore.WasmCloud.Runtime.Server.invoke_actor(rtpid, aref, "HttpServer.HandleRequest", <<>>)
```
you'll see my debug output (`HEY THERE!`) and then you'll see it panic because of a deserialization failure. If you want to see it work all the way through, you'll need to create a map, pipe it through msgpax, and then send it in.

host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
autodidaddict: 👋

autodidaddict:
status - load echo raw, start it, then use wash to call it via rpc...
```
HEY THERE
         16:12:47.140 [error] actor_id=MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 host_id=NC654BD34IS4JDKLTZJFK6BAOMQHLUCBBWWQ2OUVMZ62FELPJAFKVBFF lattice_prefix=default oci_ref= GenServer #PID<0.589.0> terminating
** (CaseClauseError) no case clause matching: :ok
```

autodidaddict:
Can invoke actor functions now. Actors that make host calls will still run into the panic that I've left in the handler.

host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+/// Called from the Elixir native wrapper which is in turn wrapped by the Wasmcloud.Runtime.Server GenServer
+#[rustler::nif(name = "start_actor")]
+pub fn start_actor<'a>(
+    env: rustler::Env<'a>,
+    runtime_resource: ResourceArc<RuntimeResource>,
+    bytes: Binary<'a>,
+) -> Result<ResourceArc<ActorResource>, rustler::Error> {
+    let actor: Actor = Actor::new(&runtime_resource.inner, bytes.as_slice())
```
rvolosatovs, line 138:
```suggestion
    let actor = Actor::new(&runtime_resource.inner, bytes.as_slice())
```

host_core/lib/host_core/wasmcloud/runtime_config.ex
```diff
@@ -0,0 +1,17 @@
+defmodule HostCore.WasmCloud.Runtime.Config do
+  @moduledoc ~S"""
+  Configures a `WasmCloud.Runtime`.
+
+  ## Options
+    * `:placeholder` - holds a place
+
+  ## Example
+      iex> _config = %HostCore.WasmCloud.Runtime.Config{}
```
brooksmtownsend, line 9: Don't forget to spec out this 😉

host_core/lib/host_core/wasmcloud/rpc_invocations.ex
```diff
@@ -0,0 +1,322 @@
...
+
+  # allow actor-to-actor calls
+  def authorize_call(%{target: {:actor, _, _}} = token) do
+    %{token | authorized: true}
+  end
+
+  # LINKS
+
+  # Built-in Providers do not have link definitions
+  # their implementations are already handled either in the NIF or in wasmCloud runtime
```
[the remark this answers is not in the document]
brooksmtownsend: And if so, can we remove the numbergen/logging provider from this host?

host_core/native/hostcore_wasmcloud_native/Cargo.toml
```diff
@@ -29,6 +34,6 @@ chrono-humanize = "0.2.1"
 chrono = "0.4.19"
 tokio-stream = "0.1"
 bindle = { version = "0.9", default-features = false, features = ["client", "caching", "rustls-tls"] }
-async-trait = "0.1"
 reqwest = { version = "0.11", default-features = false, features = ["rustls-tls"] }
-nats = "0.23.1"
+nats = "0.24.0"
```
brooksmtownsend, line 38: We should probably use async-nats now 😢 Will open up another issue

host_core/native/hostcore_wasmcloud_native/rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly"
```
brooksmtownsend, line 2: Let's note in the README that nightly is required to compile t his

host_core/test/e2e/kvcounter_test.exs
```diff
@@ -230,10 +235,10 @@ defmodule HostCore.E2E.KVCounterTest do
     Process.sleep(300)
 
     {:ok, _okay} = HTTPoison.start()
-    {:ok, resp} = request_http("http://localhost:8082/foobar", 10)
+    {:ok, resp} = request_http("http://localhost:8082/foobar", 2)
 
-    assert resp.body ==
-             "{\"error\":\"Host send error Invocation not authorized: missing capability claim for wasmcloud:keyvalue\"}"
+    # assert resp.body ==
+    #          "{\"error\":\"Host send error Invocation not authorized: missing capability claim for wasmcloud:keyvalue\"}"
```
brooksmtownsend, line 241: After the latest changes where libwasmcloud isn't checking this I think you can bring this assert back

host_core/test/host_core/actors_test.exs
```diff
@@ -470,6 +470,8 @@ defmodule HostCore.ActorsTest do
       {:ok, _pid} = ActorSupervisor.start_actor(bytes, config.host_key)
       wait_for_actor_start(evt_watcher, @pinger_key)
 
+      Process.sleep(2_000)
```
brooksmtownsend, line 473: 🙃

host_core/lib/host_core/wasmcloud/runtime.ex
```diff
@@ -0,0 +1,97 @@
+defmodule HostCore.WasmCloud.Runtime do
+  @moduledoc """
+  This module provides encapsulated access to the native NIF function calls exposed through the wasmcloud/native.ex interface
+  This runtime module supports the Runtime.Server module and should _not_ be used directly by any code other than that module. This
+  would be a private module, but Elixir doesn't allow selective access
+  """
+  @type t :: %__MODULE__{
+          resource: binary(),
+          reference: reference()
```
brooksmtownsend, line 9: So, an actor having a resource (the bytes of the actor) and a reference (reference to actor) makes sense. What's the resource of a runtime?

host_core/lib/host_core/wasmcloud/runtime_server.ex
```diff
@@ -0,0 +1,180 @@
...
+  end
+
+  # this gets called from inside the NIF to indicate that a function call has completed
+  # the `from` here is the same from (via passthrough) that came from the
+  # GenServer call to `:invoke_actor`
+  @impl true
+  def handle_info({:returned_function_call, result, from}, state) do
+    # the binary comes out of the NIF as  {:ok, vec<u8>} or {:error, vec<u8>}
+    # so we need to turn the second element from a vec<u8> into a << ...>> binary
+    bindata = elem(result, 1)
```
brooksmtownsend, line 104: Would prefer a pattern match here over an `elem` call to avoid panics if we reach this branch with an unexpected value

host_core/lib/host_core/wasmcloud/runtime_server.ex
```diff
@@ -0,0 +1,180 @@
...
+          <<>>
+
+        is_binary(bindata) && byte_size(bindata) == 0 ->
+          <<>>
+
+        true ->
+          IO.iodata_to_binary(bindata)
+      end
+
+    code = elem(result, 0)
```
brooksmtownsend, line 118: same comment re: `elem`

host_core/lib/host_core/wasmcloud/runtime_server.ex
```diff
@@ -0,0 +1,180 @@
...
+  @impl true
+  def handle_info(
+        {:invoke_callback, claims, binding, namespace, operation, payload, token},
+        {_runtime, config} = state
+      ) do
+    # This callback is invoked by the wasmcloud::Runtime's host call handler
+    payload = payload |> IO.iodata_to_binary()
+    # TODO
+    {success, return_value} =
+      try do
```
brooksmtownsend, line 135: Is there a performance impact like Java *shudders* running this inside of a `try` block?

host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
```diff
@@ -0,0 +1,239 @@
...
+        .context("failed to load actor from bytes")
+        .unwrap();
+
+    let ar = ActorResource { actor };
+
+    Ok(ResourceArc::new(ar))
+}
+
+// NOTE: wasmex uses dirtycpu here. should we use dirty IO ?
+#[rustler::nif(name = "call_actor", schedule = "DirtyCpu")]
```
brooksmtownsend, line 140: I actually don't think we need the DirtyCpu or DirtyIo here, spawning a thread likely always comes in under a millisecond and all execution is in that thread, which is separate from the erlang scheduler

brooksmtownsend (commented):
Lotsa comments and I know it's still in draft, but I wanted to drop some feedback in the meantime

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -651,70 +622,30 @@ defmodule HostCore.Actors.ActorModule do
...
-    module =
-      case :ets.lookup(:module_cache, key) do
-        [{_, cached_mod}] ->
-          cached_mod
-
-        [] ->
-          {:ok, mod} = Wasmex.Module.compile(bytes)
-          :ets.insert(:module_cache, {key, mod})
-          mod
-      end
```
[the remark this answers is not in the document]
autodidaddict: the nif is holding onto the precompiled module by way of the `ActorReference`, but yeah we don't need to store this stuff in ETS because it's already stored via Rust.

host_core/lib/host_core/actors/actor_module.ex
```diff
@@ -651,70 +622,30 @@ defmodule HostCore.Actors.ActorModule do
...
+    rtpid = HostCore.Vhost.VirtualHost.get_runtime(pid)
 
+    case HostCore.WasmCloud.Runtime.Server.precompile_actor(rtpid, bytes) do
+      {:ok, aref} ->
+        iid = UUID.uuid4()
+        ClaimsManager.put_claims(host_id, lattice_prefix, claims)
+        ActorRpcSupervisor.start_or_reuse_consumer_supervisor(lattice_prefix, claims)
         publish_oci_map(host_id, lattice_prefix, oci, claims.public_key)
-        {:ok, agent}
+        publish_actor_started(host_id, lattice_prefix, claims, "n/a", iid, oci, annotations)
```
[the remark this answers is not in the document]
autodidaddict: I don't think we do... I can't reproduce the case that originally required us to have it.

host_core/lib/host_core/actors/actor_supervisor.ex
```diff
@@ -328,6 +283,49 @@ defmodule HostCore.Actors.ActorSupervisor do
...
+        )
+
+        # Each spawned function is a new process, therefore a new root trace
+        # this is why we pass the span context so all child updates roll up
+        # to the current trace
+        Enum.each(targets, fn pid ->
+          ActorModule.live_update(
+            config,
+            pid,
+            IO.iodata_to_binary(bytes),
```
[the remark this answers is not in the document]
autodidaddict: this is old code I didn't change... but the difference is that one of these comes from a File.read! which is already binary, and the other source is a vec<u8> which comes up as a list.

host_core/lib/host_core/wasmcloud/runtime.ex
autodidaddict: the resource is actually the resource arc that comes from the NIF:
```
  let rt = WcRuntime::from_host_handler(host_handler)
        .context("failed to construct runtime")
        .map_err(|e| Error::Term(Box::new(e.to_string())))?;

    let resource = ResourceArc::new(RuntimeResource { inner: rt });
    Ok(resource)
```

The use of the binary here is an esoteric incantation required to keep the various compilers from inlining and removing data

host_core/lib/host_core/wasmcloud/runtime_server.ex
autodidaddict: try blocks are "free" in elixir

host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
autodidaddict: https://github.com/tessi/wasmex/blob/6594e2c1d592c0c58f72200d4eb2285b33c0f327/native/wasmex/src/instance.rs#L102-L103

brooksmtownsend: Never mind then 😄 I do think in principle DirtyIo makes sense here, I'd be interested in experimenting with that value since our Wasms usually take longer waiting for responses for NATS messages then they do computing things

autodidaddict: yeah it's definitely something to spike/benchmark on

host_core/lib/host_core/wasmcloud/rpc_invocations.ex
```diff
@@ -0,0 +1,322 @@
...
+
+  # default behavior is to allow actor-to-actor calls
+  def verify_link(%{target: {:actor, _, _}} = token) do
+    %{token | verified: true}
+  end
+
+  def verify_link(
```
brooksmtownsend, line 91: `verify_link` needs a branch that can handle the target not being found in `identify_target`, or else we get a nasty crash like 
```
            ** (FunctionClauseError) no function clause matching in HostCore.WasmCloud.RpcInvocations.verify_link/1
                (host_core 0.62.0) lib/host_core/wasmcloud/rpc_invocations.ex:87: HostCore.WasmCloud.RpcInvocations.verify_link(%{<blahblahtokenstuff>, error: "Could not identify suitable target for invocation", 
```

brooksmtownsend: This is because the `target` field isn't present

brooksmtownsend (changes requested):

brooksmtownsend (approved):
Things I see that don't need to block this big ol PR as it is today and are worth a followup in a smaller, future PR as we improve release candidates
- [ ] I see some warnings from unused Elixir modules / Rust imports we could clean up. Unsure if clippy needs satisfying too.
- [ ] I've seen errors about invalid module hashes with wascap, though the test fixtures seem to load the actors just fine. Need to ensure that we can properly start actors from older OCI references and things
- [ ] Would love to review some of the newest genservers again for any possible blocking opportunities as we test this out with Petclinic and gradually more complicated examples
- [ ] Have we tried to instantiate multiple virtual hosts with this model? I know that's experimental functionality but it's there

All that being said though, integration tests look good and code looks good to me

autodidaddict:
Yes we should definitely capture these as quick follow-up issues

rvolosatovs (approved):
I have seen this working, hence approving. There's definitely some work to do in follow-ups

PR #582, opened 2023-03-21
Bumped mix.exs in wasmcloud_host and chart

brooksmtownsend:
Simply bumping the release versions for wasmcloud_host to match host_core before cutting the first RC

### DIFF
--- a/wasmcloud_host/chart/Chart.yaml
+++ b/wasmcloud_host/chart/Chart.yaml
@@ -17,4 +17,4 @@ type: application
 
 version: 0.6.10
 
-appVersion: "0.61.0"
+appVersion: "0.62.0"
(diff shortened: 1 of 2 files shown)


wasmcloud_host/chart/Chart.yaml
```diff
@@ -17,4 +17,4 @@ type: application
 
 version: 0.6.10
 
-appVersion: "0.61.0"
+appVersion: "0.62.0"
```
connorsmith256, line 20: I think we're supposed to wait to bump the chart until after the host has been released

brooksmtownsend: Let me add a comment in the next one, Taylor said you can bump this but not the chart version

PR #584, opened 2023-03-23
derive version from GITHUB_REF rather than mix.exs

connorsmith256:
## Feature or Problem
We currently derive `app-version` from mix.exs. This works for "real" releases, but not for RCs or other test tags. Instead we should derive `app-version` from the tag name. This should be accessible via `GITHUB_REF`, after we trim the leading `/refs/tags/`

(Bash & GHA experts should double-check this syntax. I'm not confident in it)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -271,7 +271,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Determine version
-        run: echo "app-version=${GITHUB_REF#refs/tags/}" > $GITHUB_ENV
+        run: echo "app-version=$GITHUB_REF_NAME" > $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -271,7 +271,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Determine version
-        run: echo "app-version=$(grep '@app_vsn "' ${{env.working-directory}}/mix.exs | cut -d '"' -f2)" > $GITHUB_ENV
+        run: echo "app-version=${GITHUB_REF#refs/tags/}" > $GITHUB_ENV
```
brooksmtownsend, line 274: Do you just want it to be the tag? GITHUB_REF will be something like `refs/tags/v0.62.0-rc.1` but GITHUB_REF_NAME will be `v0.62.0-rc.1`

Issue #598, opened 2023-03-30
[FEATURE] Hide invocation event warnings in the dashboard event handler


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
Often when running wasmCloud, new users can see these warnings:
```
(com.wasmcloud.lattice.invocation_failed) received but was ignored: No handler defined"}
(com.wasmcloud.lattice.invocation_succeeded) received but was ignored: No handler defined"}
```

These are confusing, and can even point the finger at the user to say that they defined their handler wrong. This is **not** the error, we simply have an event listener in the wasmCloud dashboard that ignores these events as we don't need the information.

**Describe the solution you'd like**
The wasmCloud dashboard should have a special case for `invocation_succeeded` and `invocation_failed` to ignore them _without_ a log message.

**Describe alternatives you've considered**
We could handle these events directly, but really there's nothing to do with them.

**Additional context**
Here's the event handler fallthrough function: https://github.com/wasmCloud/wasmcloud-otp/blob/main/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex#L426-L438

We should keep that fallthrough and add two additional matching functions looking for the above events so that the invocation-related event's don't fallthrough

Issue #571, opened 2023-03-15
[BUG] Hosts do not auto discover themselves in the washboard


brooksmtownsend:
**Describe the bug**

I'm running a few wasmCloud hosts connected in the same lattice but the only one that reports to me is my local host.

![image](https://user-images.githubusercontent.com/12040685/225367932-4cd6d257-c4f9-4219-8fc5-c860c7514432.png)

**To Reproduce**

1. Start NATS
1. Start a host with `WASMCLOUD_DASHBOARD_PORT=4000 iex -s mix phx.server`
1. Start a host with `WASMCLOUD_DASHBOARD_PORT=4001 iex -s mix phx.server`
2. Note that on both dashboards you can only see one host

**Expected behavior**

Each host should auto discover the other one either by a host_started event or by the heartbeat and show up as remote hosts in the dashboard view.

**Environment (please complete the following information)**
 - OS: [e.g. Linux, MacOS, Windows] Linux / Macos
 - Arch: [e.g. x86_64, aarch64] aarch64
 - wasmCloud Version: [e.g. 0.51.0] v0.60.0

PR #600, opened 2023-03-30
Fix/washboard events

connorsmith256:
## Feature or Problem
This PR fixes two issues:
- it adds a NATS subscription on `wasmbus.evt.{lattice}` so the washboard can receive events from other hosts, namely the host started/stopped/heartbeat events
- adds missing `process_event` handlers for events. Some of these are no-ops, and some needed logic (the host started/stopped events)

## Related Issues
Fixes https://github.com/wasmCloud/wasmcloud-otp/issues/571
Fixes https://github.com/wasmCloud/wasmcloud-otp/issues/598

## Release Information
Next (preferably v0.62)

## Consumer Impact
The washboard is usable for lattices with multiple hosts in them once again :) 

### Manual Verification
Tested manually with two hosts, e.g.:

In one terminal:
```
WASMCLOUD_CLUSTER_SEED=FOO WASMCLOUD_CLUSTER_ISSUERS=BAR PORT=4000 WASMCLOUD_DASHBOARD_PORT=4000 RELEASE_NODE=first-host make run
```

In another:
```
WASMCLOUD_CLUSTER_SEED=FOO WASMCLOUD_CLUSTER_ISSUERS=BAR PORT=4001 WASMCLOUD_DASHBOARD_PORT=4001 RELEASE_NODE=second-host make run
```

I verified that when the second host starts, the first host's washboard immediately recognizes the second host. The second host's washboard will recognize the first host when it receives a heartbeat from the host. I also verified that stopping one host is correctly processed by the other host's washboard state

### DIFF
--- a/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
+++ b/wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
@@ -139,29 +139,23 @@ defmodule WasmcloudHost.Lattice.StateMonitor do
         state
       ) do
     state =
-      if String.starts_with?(topic, "wasmbus.evt.") do
-        case Cloudevents.from_json(body) do
-          {:ok, event} ->
-            {pk, _pid, _prefix} = WasmcloudHost.Application.first_host()
+      with true <- String.starts_with?(topic, "wasmbus.evt."),
+           {:ok, event = %Cloudevents.Format.V_1_0.Event{source: source}} <-
+             Cloudevents.from_json(body),
+           {pk, _pid, _prefix} <- WasmcloudHost.Application.first_host(),
+           true <- source != pk do
+        process_event(state, event)
+      else
+        # ignoring bad topic or event from this host
+        false ->
+          state
 
-            case event do
-              # process NATS events from other hosts
-              %Cloudevents.Format.V_1_0.Event{
-                source: source
-              }
-              when source != pk ->
-                process_event(state, event)
+        {:error, _} ->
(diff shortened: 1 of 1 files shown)


autodidaddict:
I'm assuming the literal value `BAR` doesn't work for the issuers value? :)

wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
```diff
@@ -113,9 +116,53 @@ defmodule WasmcloudHost.Lattice.StateMonitor do
...
+
+  # handle cloud events from NATS
+  @impl true
+  def handle_info(
+        {:msg, %{body: body, topic: topic}},
+        state
+      ) do
+    state =
+      if String.starts_with?(topic, "wasmbus.evt.") do
+        case Cloudevents.from_json(body) do
```
autodidaddict, line 143: tiny nit: we could probably clean this up with a with statement like:

```elixir
with {:ok, event} <- Cloudevents.from_json(body),
        {pk, _pid, _prefix} <- WasmcloudHost.Application.first_host() do
        case event ...
 end 
```

autodidaddict (dismissed):

wasmcloud_host/lib/wasmcloud_host/lattice/state_monitor.ex
connorsmith256: Oo yeah good call

autodidaddict (approved):

PR #603, opened 2023-04-04
strip leading v from tags

connorsmith256:
## Feature or Problem
@brooksmtownsend noticed we published v0.62.0 of the host with the leading `v` included in the tag name. That goes against the convention we've used elsewhere, so I've removed it

## Release Information
Next

## Consumer Impact
N/A

## Testing
Tested with bash:
```bash
❯ GITHUB_REF_NAME=v0.62.0

❯ echo ${GITHUB_REF_NAME#v}
0.62.0

❯ echo ${GITHUB_REF_NAME#z}
v0.62.0
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -132,7 +132,7 @@ jobs:
   #   steps:
   #     - uses: actions/checkout@v3
   #     - name: Determine version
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
Thanks for this! Will be able to test with v0.62.1

PR #605, opened 2023-04-07
feat: update to WIT-enabled wasmCloud

rvolosatovs:
## Feature or Problem

Update to NIF to https://github.com/wasmCloud/wasmCloud/pull/300

## Related Issues

https://github.com/wasmCloud/wasmCloud/pull/300

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/host_core/native/hostcore_wasmcloud_native/Cargo.toml
+++ b/host_core/native/hostcore_wasmcloud_native/Cargo.toml
@@ -10,24 +10,23 @@ path = "src/lib.rs"
 crate-type = ["dylib"]
 
 [dependencies]
-wasmcloud = { git = "https://github.com/wasmcloud/wasmcloud" }
+wasmcloud-host = { git = "https://github.com/wasmcloud/wasmcloud" }
 rustler = "0.27.0"
 lazy_static = "1.0"
 async-trait = "0.1.66"
 serde = {version = "1.0.126", features = ["derive"] }
 serde_bytes = "0.11.5"
-nkeys = "0.2.0"
+nkeys = "0.3.0"
 futures = "0.3.27"
-wascap = "0.10.1"
+wascap = "0.11.0"
 ring = "0.16.20"
-rand = "0.8.5"
 log = "0.4.17"
 uuid = {version = "1.2.1", features = ["v4"]}
 data-encoding = "2.3.2"
 rmp = "0.8.10"
 rmp-serde = "1.0.0"
 oci-distribution = { version = "0.9.1", default-features = false, features = ["rustls-tls"] }
-provider-archive = "0.6.0"
+provider-archive = "0.7.0"
 tokio = {version = "1.26.0", features = ["rt", "rt-multi-thread"] }
 once_cell = "1.2.0"
 chrono-humanize = "0.2.1"
--- a/host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
+++ b/host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
@@ -1,35 +1,45 @@
+use std::ops::Deref;
+use std::sync::{Arc, Condvar, Mutex};
+
+use crate::{
+    atoms,
+    environment::{CallbackToken, CallbackTokenResource},
+};
+
 use anyhow::{self, bail, Context};
 use async_trait::async_trait;
 use log::{error, trace};
-use rand::{thread_rng, Rng, RngCore};
-
-use crate::environment::CallbackToken;
 use rustler::{
     env::{OwnedEnv, SavedTerm},
     resource::ResourceArc,
     types::tuple::make_tuple,
     Binary, Encoder, Env, Error, LocalPid, NifResult, Term,
 };
-
-use std::sync::{Condvar, Mutex};
 use wascap::jwt;
-use wasmcloud::{
-    capability, logging, numbergen, Actor, Handle, HostInvocation, LoggingInvocation,
-    NumbergenInvocation, Runtime as WcRuntime,
-};
-
-use crate::{atoms, environment::CallbackTokenResource};
+use wasmcloud_host::{host, logging, Actor, Runtime as WcRuntime};
 
 /// A wrapper around an instance of the wasmCloud runtime. This will be used inside a `ResourceArc` to allow
 /// Elixir to maintain a long-lived reference to it
 pub struct RuntimeResource {
-    pub inner: WcRuntime,
+    inner: WcRuntime,
+    /// Runtime process ID
+    pid: LocalPid,
 }
 
 /// A wrapper around an instance of a precompiled wasmCloud actor. This will be used inside a `ResourceArc` to allow
 /// Elixir to maintain a long-lived reference to it
+#[derive(Clone)]
 pub struct ActorResource {
-    pub actor: Actor,
+    inner: Actor,
+    runtime_pid: LocalPid,
+}
+
+impl Deref for ActorResource {
+    type Target = Actor;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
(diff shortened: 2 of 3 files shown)


host_core/native/hostcore_wasmcloud_native/Cargo.toml
```diff
@@ -10,8 +10,7 @@ path = "src/lib.rs"
 crate-type = ["dylib"]
 
 [dependencies]
-# Patched temporarily to fix TinyGo modules, branch will be removed in v0.63.0
-wasmcloud = { git = "https://github.com/wasmcloud/wasmcloud", branch = "fix/0.62-tinygo"}
+wasmcloud-host = { git = "https://github.com/rvolosatovs/wasmcloud", branch = "feat/wit"}
```
autodidaddict, line 13: Calling this wasmCloud host is confusing as there is a full elixir project in this repo called wasmCloud host (the washboard). 

Also, we shouldn't be depending on your fork.

host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
```diff
@@ -38,111 +48,109 @@ pub struct ExRuntimeConfig {
...
+    /// Runtime process ID
     pid: LocalPid,
-    #[allow(unused)]
-    host_id: String,
+    claims: jwt::Claims<jwt::Actor>,
+    call_context: Vec<u8>,
+}
+
+#[derive(Clone)]
+struct ElixirHandlerArc(Arc<ElixirHandler>);
```
autodidaddict, line 60: I know naming is hard but I'd rather not put a data type like arc as a suffix on a struct.

autodidaddict (changes requested):
LGTM but I think there are some opportunities to rename some things to be more clear and disambiguated.

Also I can't tell from mobile but make sure all the integration tests pass on these new nif changes

host_core/native/hostcore_wasmcloud_native/Cargo.toml
rvolosatovs: This is a draft and the fork is used precisely because https://github.com/wasmCloud/wasmCloud/pull/300 is not merged yet. There's nothing else to depend upon to get this up

host_core/native/hostcore_wasmcloud_native/src/wasmruntime.rs
rvolosatovs: Could you suggest a better option?

autodidaddict: It was called ElixirHandler before. The fact that the inner element is an arc shouldn't be advertised in the name.

rvolosatovs: But what would the inner struct be called then? `ElixirHandlerInner`? I really don't see how that's better.

The only option to avoid this is to store an unnamed tuple in an Arc, but that'd be very difficult to maintain and use

autodidaddict: I'm also not sure what caused the need for the arc wrapper around the elixir handler. Since I'm on a phone deferring to Taylor for thoughts on the abstraction and name

rvolosatovs: > not sure what caused the need for the arc wrapper around the elixir handler

Mainly, wasmtime/wit-bindgen design drawbacks, namely the fact that the host trait implementations take `&mut self` parameter, therefore forcing this pattern on us

autodidaddict: But shouldn't we be insulating consumers of our rust wasmCloud library from design changes and decisions of the underlying libraries? We shouldn't have to refactor the otp nif every time wasmtime makes an internal change... Isolating the host from this kind of churn is actually the primary purpose of the rust wasmCloud library.

autodidaddict: > > not sure what caused the need for the arc wrapper around the elixir handler
> 
> Mainly, wasmtime/wit-bindgen design drawbacks, namely the fact that the host trait implementations take `&mut self` parameter, therefore forcing this pattern on us

If they force this on our wasmCloud library then the wasmCloud library should not pass this along to its consumers. It is designed specifically to prevent this

rvolosatovs: Sure, eventually we should think about this, depending on the state of things in the future, right now it's way too early to build these abstractions.

autodidaddict: I disagree. Our job with libwasmcloud is to isolate consumers from wasmtime and its volatility. It's our job to come up with a more stable abstraction than the raw wasmtime interface.

autodidaddict:
Bottom line: if we have to refactor the nif every time wasmtime changes an API, then libwasmcloud is not doing its job.

rvolosatovs:
> Bottom line: if we have to refactor the nif every time wasmtime changes an API, then libwasmcloud is not doing its job.

Again, that comes in the future. We will reach this point some day, but now is not the right time to build these abstractions (just like the handler API was prematurely built on your request). It's much more dangerous and costly to build the wrong abstraction, rather than pay a minuscule cost of changing a few lines of code very infrequently (if at all). Please note that "libwasmcloud" is not even `0.1.0` yet and there will be breaking changes going forward. We will follow semantic versioning completely, but now it's way too early in the lifecycle of that project to do that.

Making sure these interfaces (WIT traits, WASI etc.) are actually usable and backwards compatible is Wasmtime/Bytecode Alliance's job, not ours.

What changes are you afraid of?

Eventually, the "Arc" trick will not be necessary anymore - the (wit-bindgen generated) traits should take `&self` rather than `&mut self` as parameter - changing this should not take more than 10 seconds with a simple search-and-replace. Removing the Arc is a possibility in that case, but not a requirement - the existing implementation will keep working just the same way it did.

Next, there's the question of WIT interfaces being updated - e.g. a new method is added. Again, this maintenance burden belongs upstream - it's upstream's job to ensure backwards compatibility, e.g. ensure there are default implementations for any new methods added.

If upstream fails it's job and introduces breaking changes, then that's the point where we can abstract these traits ourselves, i.e. instead of reexporting `wit-bindgen`-generated traits, export our own with the same name and API. At this point, nothing is broken, so I do not see a reason (on the contrary, I see reasons not to) fix it

brooksmtownsend:
@rvolosatovs looks like we've gone through a few changes but settled that this PR is something we should do, would you mind providing a status update here on if this should stay draft, be replaced by another PR, etc?

autodidaddict:
We don't get anything by doing this now, but there's a lot of work that will be useful once we do
> 
> cc @autodidaddict

Don't we get the ability for the OTP host to execute modules and components built using the latest standards?

autodidaddict:
The version of `provider-archive` should be bumped to `0.7.0`

autodidaddict:
@rvolosatovs when can this be taken out of draft and made ready to merge so we can use the new dependencies?

rvolosatovs:
> @rvolosatovs when can this be taken out of draft and made ready to merge so we can use the new dependencies?

the NIF segfaults on startup, someone has to debug and fix that, I can take a look tomorrow, but if you can figure out the issue, then please go ahead

PR #610, opened 2023-04-10
BREAKING: pass log_level in HostData

connorsmith256:
## Feature or Problem
This passes `log_level` to providers. See https://github.com/wasmCloud/weld/pull/148 for context

## Related Issues
https://github.com/wasmCloud/weld/pull/148

## Release Information
v0.63.0, since this is technically a breaking change. See below.

## Consumer Impact
See the testing section of https://github.com/wasmCloud/weld/pull/148. Notice that after this change, `RUST_LOG` is ignored.

~Note: I find it odd that the source field is called `structured_log_level`, and that Elixir logging is controlled via mix environment without structured logging enabled. I propose we make this a breaking change: rename this field to just `log_level`, proxy it through to providers, and also respect it when structured logging is not enabled~

Update: we decided the following:
- `WASMCLOUD_STRUCTURED_LOG_LEVEL` is now deprecated, but will continue to be parsed
- `WASMCLOUD_LOG_LEVEL` now controls logging for structured and unstructured logs
  - this value is proxied to providers
  - If not set, it defaults to the deprecated env var, or to the mix config level (debug) if neither is set

### Manual Verification
See testing section of https://github.com/wasmCloud/weld/pull/148


connorsmith256:
After talking with @thomastaylor312, he agreed that renaming to `log_level` for elixir logs for consistency makes sense. Does anyone disagree? He also proposed that we remain backwards-compatible for a time™️ and continue to parse the old env var name, as well as the new name (`WASMCLOUD_LOG_LEVEL`)

brooksmtownsend:
@connorsmith256 I like the idea of using `log_level` and the backwards compatible logic, just output a warning if people use the old one so that they get a notice to update

brooksmtownsend (dismissed):

brooksmtownsend (approved):

brooksmtownsend (approved):

PR #615, opened 2023-04-25
build(host_core): update `wasmcloud`

rvolosatovs:
## Feature or Problem

- update `wasmcloud` (the TinyGo fix has been merged to `main`)
- remove nightly and unstable features

## Related Issues

Both of these were merged:

https://github.com/wasmCloud/wasmCloud/pull/305
https://github.com/wasmCloud/wasmCloud/pull/302

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/host_core/native/hostcore_wasmcloud_native/Cargo.toml
+++ b/host_core/native/hostcore_wasmcloud_native/Cargo.toml
@@ -10,8 +10,7 @@ path = "src/lib.rs"
 crate-type = ["dylib"]
 
 [dependencies]
-# Patched temporarily to fix TinyGo modules, branch will be removed in v0.63.0
-wasmcloud = { git = "https://github.com/wasmcloud/wasmcloud", branch = "fix/0.62-tinygo"}
+wasmcloud = { git = "https://github.com/wasmcloud/wasmcloud" }
 rustler = "0.27.0"
 lazy_static = "1.0"
 async-trait = "0.1.66"
--- a/.github/workflows/host_core.yml
+++ b/.github/workflows/host_core.yml
@@ -25,6 +25,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      # Install Rust
+      - run: rustup toolchain install stable --profile minimal
+
       # Install erlang/OTP and elixir
       - name: Install erlang and elixir
         if: ${{ startswith(matrix.os, 'ubuntu') || startswith(matrix.os, 'windows') }}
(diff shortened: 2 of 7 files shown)


host_core/native/hostcore_wasmcloud_native/Cargo.toml
```diff
@@ -10,8 +10,7 @@ path = "src/lib.rs"
 crate-type = ["dylib"]
 
 [dependencies]
-# Patched temporarily to fix TinyGo modules, branch will be removed in v0.63.0
-wasmcloud = { git = "https://github.com/wasmcloud/wasmcloud", branch = "fix/0.62-tinygo"}
+wasmcloud = { git = "https://github.com/wasmcloud/wasmcloud" }
```
autodidaddict, line 13: should this be pinned to a hash? If not, it could potentially pull a breaking change and break a build

autodidaddict (commented):
Just a question about deps

host_core/native/hostcore_wasmcloud_native/Cargo.toml
rvolosatovs: Since `Cargo.lock` is checked in, this will not happen unless explicitly requested via an update command. See https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/specifying-dependencies.md#specifying-dependencies-from-git-repositories

So if a breaking change is introduced to `main` of https://github.com/wasmCloud/wasmcloud, after `cargo update` is executed the test should fail. I would consider that a feature and not a bug, since that simplifies maintenance, meaning that we don't need to manually copy-paste revs every time we want to update this dep.

Eventually (once https://github.com/bytecodealliance/preview2-prototyping is released to crates.io), we will be able to simply consume releases here. I believe ETA on this is end of spring

brooksmtownsend: As long as the `Cargo.lock` won't auto update, I'm cool with it. I was wondering the same, like how cargo will pull in patch releases automatically

brooksmtownsend (approved):

Issue #618, opened 2023-05-05
[FEATURE] Support optional `correlation_id` on control interface requests


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
Interacting with the wasmCloud host is almost always done through the control interface. Whether it's `wash`, the lattice controller, or just publishing on the right NATS topic, these commands are how the host can be told to start/stop actors, providers, link them, etc.

When it comes to knowing what happened in response to a command, you can either:
1. Wait for a synchronous reply from the host telling you what it did, but all commands are designed to "ack" rather than confirm the completion of work
2. Wait for an event describing what happened (subscribe to `wasmbus.evt.>` and wait for `provider_started`).

When choosing option 2, there's no real way to know that an event happened in response to **your** command. For example, if two different control clients publish a command to start the same actor, you would see two `actor_started` event but not know which one corresponded to your request. This can be especially problematic when considering [wadm](https://github.com/wasmCloud/wadm), which will often ask for many actors to be started.

**Describe the solution you'd like**
A simple way to do this, as suggested by @autodidaddict, is to support "correlation IDs". Effectively, each command that comes in through the control interface could include a unique (up to the client to determine this) identifier for a command, and every event that is published in response to that command will also include the same correlation ID. In practice, a time-sortable unique ID would be most useful here.

**Describe alternatives you've considered**
We've attempted, and mostly completed, a prototype of **wadm** that uses just annotations in order to uniquely identify events. However, there still cases where this isn't sufficient, especially when starting 10000 actors. It's very difficult to put a time frame on how long this _should_ take to complete, so this PR is most useful when combined with #617.

liamwh:
I'm new here and no elixir dev, so am uncertain if the commands are CloudEvents, but if they are, and we want to embed something for tracing into the control interfaces commands, is it an idea that we follow the [CloudEvents distributed tracing spec](https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/distributed-tracing.md) and add a [traceparent attribute](https://w3c.github.io/trace-context/#traceparent-header)?

brooksmtownsend:
@liamwh Great question, the commands themselves are not CloudEvents but the resulting events are. We could certainly follow that idea and add NATS headers (the project we use to send messages), since the control interface works with those 🤔

liamwh:
@brooksmtownsend Nice thinking! 2 remarks:

[...]

PR #622, opened 2023-05-09
[WIP] Support correlation IDs on ctl commands

brooksmtownsend:
## Feature or Problem
This PR adds in an additional, optional, field that we pull off of each control interface command that has resulting events. Notably, start/scale/update/stop actor, start/stop provider, and stop host. 

The correlation ID is then passed down through the resulting functions, and then attached to resulting events from that command. This is very useful to know when a command _you_ publish directly results in an actor starting, especially if there's more than one control client that is sending a host commands. The primary use case that we envisioned in this PR is to have [wadm](https://github.com/wasmCloud/wadm) be able to publish a command from one of its scalers, and then monitor events on the lattice that correspond to a particular command in order to know when a specific event happens because of a ctl request.

## Related Issues
#618

## Release Information
`0.63.0`

## Consumer Impact
This is backwards compatible with older control interface clients so no changes will be necessary for this host version.

## Testing
WIP. Still some combing through code needed, though tests all pass locally.

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/host_core/lib/host_core/actors/actor_module.ex
+++ b/host_core/lib/host_core/actors/actor_module.ex
@@ -133,8 +133,8 @@ defmodule HostCore.Actors.ActorModule do
   Halts the actor module corresponding to the supplied process ID. This will attempt a graceful termination
   and will try and emit an `actor_stopped` event.
   """
-  def halt(pid) do
-    if Process.alive?(pid), do: GenServer.call(pid, :halt_and_cleanup)
+  def halt(pid, correlation_id) do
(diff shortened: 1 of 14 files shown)


brooksmtownsend:
Closing, since I don't think we need this right now.

PR #632, opened 2023-05-24
fix(ci): Builds docker images for both supported arches

thomastaylor312:
## Feature or Problem

The current 0.63 builds only have the arm64 architecture available for docker images

## Release Information

Next. However, if we can find a way to rebuild the 0.63.1 images using this, that would be nice

## Testing

This is being tested with a run of the pipeline before changing back the vars


thomastaylor312:
Closing this out and pushing to a branch in the repo so I can test the push properly

PR #633, opened 2023-05-24
fix(ci): Builds docker images for both supported arches

thomastaylor312:
Feature or Problem
The current 0.63 builds only have the arm64 architecture available for docker images

Release Information
Next. However, if we can find a way to rebuild the 0.63.1 images using this, that would be nice

Testing
This is being tested with a run of the pipeline before changing back the vars

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -333,25 +333,12 @@ jobs:
     # ```
     # Error: buildx failed with: ERROR: invalid tag "wasmcloud.azurecr.io/wasmcloud_host:616/merge": invalid reference format
     # ```
-    if: startswith(github.ref, 'refs/tags/') # Only run on tag push
+    if: startswith(github.ref, 'refs/tags/')
     strategy:
       matrix:
-        config:
-        - app-name: host_core
-          target: aarch64_linux_gnu
-          platform: linux/arm64
-
-        - app-name: host_core
-          target: x86_64_linux_gnu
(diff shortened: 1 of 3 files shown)


thomastaylor312:
I tested this on an arm device and an x86 device. You can see that it actually starts to run now with 

```
docker run --rm -it wasmcloud/wasmcloud_host:taylor-test
docker run --rm -it wasmcloud/host_core:taylor-test
```

.github/workflows/release.yml
```diff
@@ -386,25 +373,32 @@ jobs:
...
 
       - name: Build and release docker image
         uses: docker/build-push-action@v3
         with:
           builder: ${{ steps.buildx-builder.outputs.name }}
-          push: ${{ startswith(github.ref, 'refs/tags/') }}
-          context: ${{ matrix.config.app-name }}
-          file: ${{ matrix.config.app-name }}/Dockerfile
-          platforms: ${{ matrix.config.platform }}
+          push: true #${{ startswith(github.ref, 'refs/tags/') }}
```
ricochet, line 390: Need to remove now that it works

ricochet (changes requested):

.github/workflows/release.yml
thomastaylor312: Thanks for catching. Now fixed

connorsmith256 (approved):

ricochet (approved):

PR #636, opened 2023-05-31
tiny logging improvement

autodidaddict:
* Adds a few extra bits of very useful information to the debug logs.
* Adds a couple of fallthrough handlers to avoid the function pattern mismatch type of errors

Note that there were reports of the metadata cache loader "stopping". I can't reproduce this and, more importantly, the way Gnat works it traps the call to `request` in an exception handler so the worst case scenario for a Gnat consumer supervisor would be an error log dump. If it stops getting bucket updates, then that may be caused by something else.

Regardless, the changes made here add more useful troubleshooting information and certainly don't make things worse.

### DIFF
--- a/host_core/lib/host_core/jetstream/metadata_cache_loader.ex
+++ b/host_core/lib/host_core/jetstream/metadata_cache_loader.ex
@@ -63,7 +63,7 @@ defmodule HostCore.Jetstream.MetadataCacheLoader do
 
       {:error, e} ->
         Logger.error(
-          "Key added (claioms) handler failed. Failed to deserialize claims from metadata cache: #{inspect(e)}"
+          "Key added (claims) handler failed. Failed to deserialize claims from metadata cache: #{inspect(e)}"
         )
     end
   end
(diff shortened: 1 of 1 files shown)


host_core/lib/host_core/jetstream/metadata_cache_loader.ex
```diff
@@ -56,7 +62,9 @@ defmodule HostCore.Jetstream.MetadataCacheLoader do
         ClaimsManager.cache_claims(lattice_prefix, public_key, claims)
 
       {:error, e} ->
-        Logger.error("Failed to deserialize claims from metadata cache: #{inspect(e)}")
+        Logger.error(
+          "Key added (claioms) handler failed. Failed to deserialize claims from metadata cache: #{inspect(e)}"
```
connorsmith256, line 66: Nit: typo

autodidaddict: How dare you insult my `claioms`

connorsmith256 (approved):

